PR #3, opened 2022-07-31
Products field + config to use the modules component(s)

BobWez98:


BobWez98:
Closed because of https://github.com/rapidez/statamic/compare/master...BobWez98:feature/product-sync?expand=1

PR #4, opened 2022-08-08
Feature/globals

BobWez98:


src/RapidezStatamicServiceProvider.php
```diff
@@ -4,11 +4,30 @@
...
     {
         Eventy::addFilter('routes', fn ($routes) => array_merge($routes ?: [], [__DIR__.'/../routes/fallback.php']));
+
+        return $this;
+    }
+
+    public function bootGlobals() : self
+    {
+        View::composer('*', function ($view) {
+            GlobalSet::all()->each(fn ($set) => $view->with($set->handle(), $set->inCurrentSite()));
```
BobWez98, line 28: This adds the default functionality of statamic where all the globals are available everywhere

routes/fallback.php
```diff
@@ -3,8 +3,8 @@
 
 if ($entry = Data::findByRequestUrl(request()->url())) {
     if (View::exists($entry->get('template'))) {
-        echo view($entry->get('template'), $entry->data());
+        echo view($entry->get('template'), $entry);
     } else if (View::exists(str($entry->structure()->handle())->singular()->toString())) {
-        echo view(str($entry->structure()->handle())->singular()->toString(), $entry->data());
+        echo view(str($entry->structure()->handle())->singular()->toString(), $entry);
```
BobWez98, line 8: `$entry->data()` grabbed all the raw data without parsing. passing $entry to the view is exactly the same, but strings as `entry::uuid-asdf-fdjakl` will be parsed to `/url`.

PR #7, opened 2022-10-25
Refactored product sync

kevinmeijer97:


src/RapidezStatamicServiceProvider.php
```diff
@@ -39,4 +67,36 @@ public function bootGlobals() : self
...
+        $this->publishes([
+            __DIR__.'/../resources/content/collections/products.yaml' => base_path('content/collections/products.yaml'),
+        ], 'rapidez-collections');
+
+        $this->publishes([
+            __DIR__.'/../resources/blueprints/collections/stores/stores.yaml' => resource_path('blueprints/collections/stores/stores.yaml'),
+        ], 'rapidez-collections');
+
+        $this->publishes([
+            __DIR__.'/../resources/content/collections/stores.yaml' => base_path('content/collections/stores.yaml'),
```
royduin, line 97: These can be bundled into one `publishes()`

src/RapidezStatamicServiceProvider.php
```diff
@@ -4,27 +4,55 @@
...
+        ]);
+
+        return $this;
+    }
+
+    public function bootComposers() : self
+    {
+        View::composer('rapidez::product.overview', function (RenderedView $view) {
+            $entry = Entry::whereCollection('products')->where('sku', config('frontend.product')['sku'])->first();
+            $view->with('content', $entry);
```
royduin, line 53: Is this always wanted/needed? Can imagine someone only wants the products in Statemic so it can be related to other collections instead of replenishing the product with fields from Statemic.

src/Actions/ImportProducts.php
```diff
@@ -0,0 +1,38 @@
...
+    ): void
+    {
+        $productModel = config('rapidez.models.product');
+        $storeModel = config('rapidez.models.store');
+        $storeModels = $store ? $storeModel::where('store_id', $store)->get() : $storeModel::all();
+
+        foreach($storeModels as $storeModel) {
+            config()->set('rapidez.store', $storeModel->store_id);
+
+            $products = $productModel::selectAttributes(['sku', 'name', 'url_key'])->where('catalog_product_flat_'.config('rapidez.store').'.visibility', 4)->get();
```
royduin, line 29: Sure only `visibility` 4? What about 2 and 3?

src/Actions/ImportProducts.php
```diff
@@ -0,0 +1,38 @@
...
+            $storeModel = Entry::whereCollection('stores')->where('store_id', $storeModel->store_id)->first();
+
+            $products = $this->createProducts->create($products, $storeModel->id());
+            $this->deleteProducts->delete($products, $storeModel->id);
```
royduin, line 33: We're importing and so creating products and directly after deleting them?

src/Actions/CreateStores.php
```diff
@@ -0,0 +1,32 @@
...
+            'store_code' => $store->code,
+            'store_id' => $store->store_id,
+            'title' => $store->name,
+            'locale' => $this->getLocaleFromStore($store->store_id),
+        ])->each(fn ($store) => Entry::updateOrCreate($store, 'stores', 'store_id'));
+    }
+
+    public function getLocaleFromStore(int $storeId): string
+    {
+        $store = DB::table('core_config_data')
```
royduin, line 23: We could use the config model here right? So we don't have to handle the fallback ourself

src/Actions/ImportProducts.php
royduin, line 29: How efficient is this with a lot of products? Maybe chunk the results like the Rapidez product indexer does?

royduin (changes requested):
And don't forget to add more info to the readme

src/Actions/ImportProducts.php
kevinmeijer97: Yes, we're passing the imported products to the delete function where we do a compare to check if the products in Statamic and the imported (Magento) products are the same. If there are any products in Statamic that do not exist in Magento, then we delete those products from Statamic. Maybe the naming was a bit confusing, changed it to deleteOldProducts().

PR #8, opened 2022-11-01
Making globals more global

Jade-GG:
This gives content pages that would previously not have access to globals, access to globals from statamic.

I'm personally unsure why the previous version didn't apply to content pages that extend rapidez::layouts.app. This just kinda forces it to work, and there's almost certainly a cleaner way.

### DIFF
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -63,7 +63,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
             });
         }
 
-        View::composer('rapidez::layouts.app', function ($view) {
+        View::composer('*', function ($view) {
             foreach (GlobalSet::all() as $set) {
                 foreach ($set->localizations() as $locale => $variables) {
                     if ($locale == Site::current()->handle()) {
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -61,7 +61,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
 
     private function getGlobals() : array
     {
-        return Cache::rememberForever('statamic-globals', function() {
+        return Cache::rememberForever('statamic-globals-'.Site::current()->handle(), function() {
             foreach (GlobalSet::all() as $set) {
                 foreach ($set->localizations() as $locale => $variables) {
                     if ($locale == Site::current()->handle()) {
@@ -92,7 +92,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
     public function bootListeners() : self
     {
         Event::listen([GlobalSetSaved::class, GlobalSetDeleted::class], function() {
-            Cache::forget("statamic-globals");
+            Cache::forget('statamic-globals-'.Site::current()->handle());
         });
 
         return $this;
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -15,6 +15,10 @@ use Illuminate\Support\Facades\View;
 use Illuminate\View\View as RenderedView;
 use Statamic\Facades\GlobalSet;
 use Statamic\Facades\Site;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\Event;
+use Statamic\Events\GlobalSetSaved;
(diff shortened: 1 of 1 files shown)


royduin:
This is bad for performance, cache it with for example the `array` store within the function:
```
Cache::store('array')->rememberForever('strapi-globals', function () {
    //
})
```

Or maybe better, something like: https://laracasts.com/discuss/channels/laravel/executing-a-view-composer-only-once

Jade-GG:
I have added a rememberForever cache and hooked into the GlobalSet* statamic events to clear the cache when necessary :)

src/RapidezStatamicServiceProvider.php
```diff
@@ -54,6 +59,20 @@ public function bootCommands() : self
         return $this;
     }
 
+    private function getGlobals() : array
+    {
+        return Cache::rememberForever('statamic-globals', function() {
```
BobWez98, line 64: Maybe add the locale to the cache key here, so every locale will get their own cache. Otherwise if you have a EN and ES store they both get the same.

src/RapidezStatamicServiceProvider.php
```diff
@@ -63,16 +82,17 @@ public function bootComposers() : self
...
 
-        View::composer('rapidez::layouts.app', function ($view) {
-            foreach (GlobalSet::all() as $set) {
-                foreach ($set->localizations() as $locale => $variables) {
-                    if ($locale == Site::current()->handle()) {
-                        $data[$set->handle()] = $variables;
-                    }
-                }
-            }
+        View::composer('rapidez::*', function ($view) {
```
BobWez98, line 85: I think we need globals everywhere, not only in the rapidez namespace. This will only be available in `resources/views/vendor/rapidez` but new views in a project won't be added here.

BobWez98 (changes requested):

src/RapidezStatamicServiceProvider.php
Jade-GG: I had it as * originally, but it turns out the statamic globals page breaks with that because it happens to conflict with its own variable called $globals. And there doesn't seem to be a clean way to exclude specific views from a composer.

Perhaps changing the variable name would be a solution but that'd be a breaking change.

BobWez98 (approved):

kevinmeijer97 (approved):

royduin:
Did anyone look at https://laracasts.com/discuss/channels/laravel/executing-a-view-composer-only-once? This vs that? @BobWez98 @kevinmeijer97 @indykoning

BobWez98:
> Did anyone look at https://laracasts.com/discuss/channels/laravel/executing-a-view-composer-only-once? This vs that? @BobWez98 @kevinmeijer97 @indykoning

Ah yes, i think a singleton would be an improvement in performance. So let's do that :)

Jade-GG:
I feel like the singleton approach would be slightly more complicated w.r.t. updating on change (and maybe a little overkill here) but I won't stop you from making it happen :)

Meanwhile I've added a little fix to allow the globals to be everywhere without Statamic breaking.

Another option I just found is [View::share](https://laravel.com/docs/9.x/views#sharing-data-with-all-views) but I don't think this would let you get around the statamic breaking issue and it seems to have an order-of-operations issue in this case.

royduin:
If you could create another PR so we can choose which we like most?

indykoning:
> Did anyone look at [laracasts.com/discuss/channels/laravel/executing-a-view-composer-only-once](https://laracasts.com/discuss/channels/laravel/executing-a-view-composer-only-once)? This vs that? @BobWez98 @kevinmeijer97 @indykoning

On the point of cache vs singleton, and this is gonna get nitpicky on performance;
The singleton will be more performant within the same request considering the values are remembered within an array in the current request.
While `Cache::rememberForever()` (unless you're using the array driver) still needs to do the round trip to your caching provider (file, database, redis, etc.) which while not needing a lot of time, do need extra time over direct array access.

If even a single instance of the loop has performance impact you could combine the singleton and trying to fetch from cache.
This way it first tries to access the data from the array, if it doesnt have it, get it from cache and if that doesn't have it build it.
That would be the most performant way but might be overkill.
Just keep in mind that this gets called A LOT during the rendering of a template so small performance issues can blow up

royduin:
Closed in favor of #10

PR #9, opened 2022-11-04
Added Magento products and categories imports

kevinmeijer97:


resources/content/collections/categories.yaml
```diff
@@ -0,0 +1,13 @@
...
+layout: layout
+revisions: false
+slugs: false
+sort_dir: asc
+date_behavior:
+  past: public
+  future: private
+structure:
+  root: false
+  slugs: true
```
BobWez98, line 13: above there's `slugs: false`, what's the diference?

src/Actions/Categories/CreateCategories.php
```diff
@@ -0,0 +1,27 @@
...
+            return new Collection();
+        }
+
+        return $categories->filter(fn ($category) => !blank($category->name) && !blank($category->url_path) && !blank($category->entity_id))->map(fn ($category) => [
+            'magento_entity_id' => $category->entity_id,
+            'magento_parent_id' => $category->parent_id ?? '',
+            'title' => $category->name,
+            'slug' => Str::replace('/', '-', $category->url_path),
+            'path' => $category->url_path ?? '',
+        ])->each(function ($category) use ($storeCode) {
```
BobWez98, line 23: if you use a shorhand function, you can remove the use.

resources/content/collections/categories.yaml
kevinmeijer97: This option shows the slug in the tree view in the CP. This was not necessary for the default view.

src/Actions/Categories/CreateCategories.php
kevinmeijer97: Gotcha, gonna check the rest as well!

src/Http/Controllers/StatamicRewriteController.php
```diff
@@ -5,17 +5,25 @@
...
 
 class StatamicRewriteController
 {
     public function __invoke(Request $request)
     {
+        $storeCode = config()->get('rapidez.store_code');
+
         $entry = Entry::query()
             ->where('collection', 'pages')
+            ->where('locale', $storeCode)
```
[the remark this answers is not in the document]
kevinmeijer97: Locale is not actually the locale on an entry, instead it's the site. A bit confusing..

PR #10, opened 2022-11-07
Globals everywhere via singleton

Jade-GG:
Alternative version that uses a singleton, reducing the amount of Cache calls from many (very many) per request to just one

### DIFF
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -5,11 +5,14 @@ namespace Rapidez\Statamic;
 use Illuminate\Support\ServiceProvider;
 use TorMorten\Eventy\Facades\Eventy;
 
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\Event;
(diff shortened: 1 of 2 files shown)


Jade-GG:
@indykoning @royduin @BobWez98 @kevinmeijer97 this is the alternative version that uses a singleton for more efficiency.

src/RapidezStatamicServiceProvider.php
```diff
@@ -19,14 +19,17 @@
 
 class RapidezStatamicServiceProvider extends ServiceProvider
 {
+    $this->app->singleton(StatamicDataComposer::class);
```
royduin, line 22: Uh? This should be in a method somewhere?

royduin:
Shouldn't be there a `compose()` method? See: https://laravel.com/docs/master/views#view-composers

src/RapidezStatamicServiceProvider.php
Jade-GG: whoops

PR #14, opened 2022-11-28
Only index indexable products

kevinmeijer97:


src/Actions/Products/ImportProducts.php
```diff
@@ -30,7 +31,13 @@ public function import(
             config()->set('rapidez.store', $siteAttributes['magento_store_id']);
 
             $productSkus = collect();
-            $productQuery = $productModel::selectAttributes(['sku', 'name', 'url_key']);
+            $childIds = DB::table('catalog_product_super_link')->select(['product_id'])->get();
```
BobWez98, line 34: If you add `->pluck('product_id')` here, you don't need the line below.

src/Actions/Products/ImportProducts.php
```diff
@@ -30,7 +31,13 @@ public function import(
...
+            $childIds = $childIds->map(fn ($item) => $item->product_id);
+            $flat = (new $productModel())->getTable();
+            $productQuery = $productModel::selectOnlyIndexable()
+                ->where($flat . '.type_id', 'configurable')
+                ->orWhereNotIn($flat . '.entity_id', $childIds->toArray())
+                ->withEventyGlobalScopes('index.product.scopes');
```
BobWez98, line 40: I don't think we need the indexer scope here, all that does is request extra data which we don't need for this sync.

BobWez98 (changes requested):

PR #15, opened 2022-12-01
Refactored routing and moved viewcomposer

BobWez98:


src/RapidezStatamicServiceProvider.php
```diff
@@ -31,26 +31,27 @@ public function boot()
...
-            ->bootListeners();
+            ->bootListeners()
+            ->bootRoutes();
     }
 
-    public function bootConfig() : self
+    public function bootRoutes() : self
     {
-        $this->mergeConfigFrom(__DIR__.'/../config/statamic.php', 'statamic');
+        if (!$this->app->routesAreCached()) {
```
BobWez98, line 41: Maybe this is something for every rapidez package. Because it is unnecessary to load routes when they're cached.

royduin: Indeed, PR bomb incoming... 😋

PR #19, opened 2023-02-06
Fixed cache not actually updating on save

Jade-GG:


indykoning:
This is what we have https://laravel.com/docs/9.x/cache#storing-items-forever for 🙂

PR #21, opened 2023-02-07
Add product visibility to product import

Jade-GG:
Also made addHidden an option

### DIFF
--- a/src/Actions/Products/ImportProducts.php
+++ b/src/Actions/Products/ImportProducts.php
@@ -20,7 +20,7 @@ class ImportProducts
     ) {
     }
 
-    public function import(?Carbon $updatedAt = null, ?string $store = null): void
+    public function import(?Carbon $updatedAt = null, ?string $store = null, ?bool $addHidden = false): void
     {
         $productModel = config('rapidez.models.product');
 
@@ -30,7 +30,7 @@ class ImportProducts
             $sites = collect([$store => Site::get($store)]);
         }
 
-        $sites->each(function ($site) use ($productModel, $updatedAt) {
+        $sites->each(function ($site) use ($productModel, $updatedAt, $addHidden) {
             $siteAttributes = $site->attributes();
 
             if (!isset($siteAttributes['magento_store_id'])) {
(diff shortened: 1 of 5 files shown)


src/Actions/Products/ImportProducts.php
```diff
@@ -90,8 +92,6 @@ public function delete(?string $store = null): void
             $productSkus = collect();
 
             $productQuery = $productModel::selectOnlyIndexable()
-                ->where($flat->qualifyColumn('type_id'), 'configurable')
-                ->orWhereNotIn($flat->qualifyColumn('entity_id'), $childIds->toArray())
```
Jade-GG, line 94: Is it necessary to add a `->when()` to keep this part in?

src/Actions/Products/ImportProducts.php
```diff
@@ -90,8 +92,6 @@ public function delete(?string $store = null): void
             $productSkus = collect();
 
             $productQuery = $productModel::selectOnlyIndexable()
```
royduin, line 94: We only need the sku and name right? Just select them explicitly instead of all indexable attributes
```suggestion
            $productQuery = $productModel::selectAttributes(['sku', 'name'])
```
Probably all the global scopes are also not needed which speeds up the query

src/Actions/Products/ImportProducts.php
```diff
@@ -89,13 +91,10 @@ public function delete(?string $store = null): void
 
             $productSkus = collect();
 
-            $productQuery = $productModel::selectOnlyIndexable()
-                ->where($flat->qualifyColumn('type_id'), 'configurable')
-                ->orWhereNotIn($flat->qualifyColumn('entity_id'), $childIds->toArray())
-                ->withEventyGlobalScopes('statamic.product.scopes');
+            $productQuery = $productModel::selectAttributes(['sku', 'name']);
```
royduin, line 94: You can keep the custom statamic one, but remove the others. Something like:
```suggestion
            $productQuery = $productModel::selectAttributes(['sku', 'name'])
                ->withoutGlobalScopes()
                ->withEventyGlobalScopes('statamic.product.scopes');
```

PR #28, opened 2023-04-25
v1

royduin:


src/RapidezStatamicServiceProvider.php
```diff
@@ -2,85 +2,85 @@
...
-        $this->bootCommands()
+        $this
             ->bootConfig()
+            ->bootViews()
             ->bootPublishables()
             ->bootComposers()
             ->bootListeners()
-            ->bootRoutes();
+            ->bootRoutes()
+            ->bootRunway();
```
Jade-GG, line 30: This is a small nitpick, but (in my opinion) if we're going to have any specific ordering of chaining here the functions below should at least be in the same order as they are called up here.

royduin: Fixed

PR #29, opened 2023-05-04
Add category collection to category pages

Jade-GG:
I had already thought this was included, but it seems like it's not.

This could do with a refactor, however because a much bigger (runway) refactor is already coming that seems unnecessary?


royduin:
This will be present in the next release

PR #34, opened 2023-06-07
Add conditionals to form via JS Driver

Jade-GG:
Also adds a simple `select` field

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -1,7 +1,3 @@
 import 'Vendor/rapidez/core/resources/js/vue'
 
-// Ensure Vue has access to global `Statamic` component
-Vue.prototype.Statamic = window.Statamic;
-
-// Create `<form-conditions>` component to be used with Vue JS driver
 Vue.component('form-conditions', () => import('./components/FormConditions.vue'));
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -5,14 +5,4 @@ import Vue from 'vue'
 Vue.prototype.Statamic = window.Statamic;
 
 // Create `<form-conditions>` component to be used with Vue JS driver
-Vue.component('form-conditions', {
-    props: ['initialData'],
-    data() {
-        return {
-            formData: this.initialData,
-        };
-    },
-    render() {
-        return this.$scopedSlots.default({ formData: this.formData });
-    },
-});
+Vue.component('form-conditions', () => import('./components/FormConditions.vue'));
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -1,0 +1,18 @@
+import '/public/vendor/statamic/frontend/js/helpers.js'
+import Vue from 'vue'
+
+// Ensure Vue has access to global `Statamic` component
+Vue.prototype.Statamic = window.Statamic;
+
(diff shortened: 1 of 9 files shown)


src/Forms/Vue.php
royduin, line 1: Maybe move it to another subdirectory just like Statamic: `Rapidez\Statamic\Forms\JsDrivers` so we know what it is.

Jade-GG: Ah you're right, I'll stick to the consistency with the Statamic version (This was the path I stole from the Vue example)

resources/views/form_fields/select.blade.php
```diff
@@ -0,0 +1,22 @@
+<x-input.select
```
royduin, line 1: This component doesn't exist, see: https://github.com/rapidez/core/tree/master/resources/views/components

resources/js/package.js
royduin, line 1: How big is this? Maybe move it to a Vue component file which will be lazy loaded? This way it's bundled and available on all pages.

resources/views/form_fields/select.blade.php
```diff
@@ -0,0 +1,22 @@
...
+    :class="$field['error'] ? 'border-red-500' : ''"
+    :required="in_array('required', $field['validate'] ?? [])"
+    v-model="formData.{{ $field['handle'] }}"
+>
+    @if (!($field['default'] ?? ''))
+        <option
+            value="null"
+            disabled
+            selected
+        >{{ $field['placeholder'] ?? '' }}</option>
```
royduin, line 14:
```suggestion
        <option value="null" disabled selected>
            {{ $field['placeholder'] ?? '' }}
        </option>
```

resources/views/form_fields/select.blade.php
```diff
@@ -0,0 +1,22 @@
...
+    @endif
+    @foreach ($field['options'] ?? [] as $option => $label)
+        <option
+            value="{{ $option }}"
+            {{ $option == ($field['default'] ?? '') ? 'selected' : '' }}
+        >{{ $label }}</option>
```
royduin, line 20:
```suggestion
        <option value="{{ $option }}" @selected($option == $field['default'] ?? '')>
            {{ $label }}
        </option>
```

resources/js/package.js
Jade-GG: It's a very small component, but I guess these small components do add up over time so it's worth to lazy load it (especially given it's only used on very specific pages)

resources/js/package.js
```diff
@@ -0,0 +1,8 @@
+import '/public/vendor/statamic/frontend/js/helpers.js'
```
royduin, line 1: This is still within the base bundle, can we move this / everything to the component?

resources/js/package.js
royduin: More for the included Statamic JS, not sure how big that is?

resources/js/package.js
Jade-GG: The helpers import is required for the `window.Statamic` function I'll try to mess with it to see if I can minimize it though

royduin: Can't we just move everything to the component?

PR #39, opened 2023-09-07
Don't use children_count

Jade-GG:


royduin:
Where is `product_count` coming from?

Jade-GG:
> Where is `product_count` coming from?

Hm, I was going to link the code where we implemented this recently but it looks like I was mistaken and this hasn't been properly implemented in the core yet... My mistake 😅

Either way, children_count should not be used for this purpose so I'll just remove the line from this PR for now.

PR #41, opened 2023-10-13
Remove ambiguous keys from query to help runway figure itself out

Jade-GG:
Runway crafts queries that don't take into account the possibility of having a query like this. For example, it assumes that the primary key is unique and doesn't need to be qualified.

By adding these subqueries we can rename the ambiguous keys before they become a problem.

### DIFF
--- a/src/Models/Brand.php
+++ b/src/Models/Brand.php
@@ -2,10 +2,10 @@
 
 namespace Rapidez\Statamic\Models;
 
-use DoubleThreeDigital\Runway\Resource;
 use DoubleThreeDigital\Runway\Traits\HasRunwayResource;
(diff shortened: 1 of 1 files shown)


indykoning:
Is the global scope needed for the runway parts? Or is it just for displaying?
We might just be able to remove the global scope when we use it in a runway context

Jade-GG:
> Is the global scope needed for the runway parts? Or is it just for displaying? We might just be able to remove the global scope when we use it in a runway context

Yeah, without the global scope we can't differentiate between brands and any other type of attribute option. We could rename the resource to `Attribute options` and remove the scope, but that'd be very weird.

PR #44, opened 2023-11-30
Move config to files in Rapidez folder

indykoning:
As the runway config entry became quite large i moved it to it's own config file, as we had done with rapidez/core

This could be considered breaking as any overwriting of the previous config location is ignored


config/rapidez/statamic/runway.php
royduin, line 1: I'm not sure about this one, what do you think @kevinmeijer97? Maybe just keep it within the `config/rapidez/statamic.php`?

kevinmeijer97: I feel like it would make more sense to keep it in the same file, because the `fetch` array in the Statamic config is also apart of the runway settings so they would have to be moved too. 
In that case i'd say keep it in the same file as it wouldn't really make sense to split it for 1 setting.

royduin:
Conflicts 😇

PR #46, opened 2023-12-05
Add product import command & add utility to start imports

FinnPaes:
This PR adds the product import command.
This also adds a utility to run the import commands from the Statamic CP.

### DIFF
--- a/src/Commands/ImportProducts.php
+++ b/src/Commands/ImportProducts.php
@@ -13,23 +13,15 @@ use Statamic\Facades\Entry;
 use Statamic\Facades\Site;
 use Symfony\Component\Console\Helper\ProgressBar;
 
-class ImportCategories extends Command
+class ImportProducts extends Command
 {
-    protected $signature = 'rapidez:statamic:import:categories {categories?* : Categories in a space separated list, by entity_id or url_key} {--all : Process all categories if none have been passed} {--site=* : Sites handles or urls to process, if none are passed it will be done for all sites (Default will be all sites)}';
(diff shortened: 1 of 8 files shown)


src/Commands/ImportProducts.php
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            Rapidez::setStore($siteAttributes['magento_store_id']);
+
+            $products = $productModel::query()
+                ->selectAttributes(['entity_id', 'sku', 'name', 'url_key'])
+                ->lazy();
+            
+            foreach ($products as $product) {
+                static::createEntry(
```
kevinmeijer97, line 48: What happens with the title when making the product content entry? Does it use the product name by default?

FinnPaes: Yes it does use the product name as title, this happens in the `RapidezStatamicServiceProvider` in the `bootListeners` function.
```php
Event::listen('rapidez-statamic:product-entry-data', fn($product) => [
     'title' => $product->name,
     'slug' => trim($product->url_key),
]);
```

src/Commands/ImportProducts.php
```diff
@@ -0,0 +1,101 @@
...
+
+            $bar->advance();
+        }
+        $bar->finish();
+
+
+        return static::SUCCESS;
+    }
+
+    protected static function createEntry(array $attributes, array $values = [])
```
indykoning, line 70: These functions are the same as the ones for the category right? 
Perhaps it's a good idea to move these to an action to be more DRY

indykoning (changes requested):
Could you update the readme with the changes for the product import 🙂

royduin (approved):

indykoning (approved):

PR #50, opened 2024-01-09
Use Eventy for import commands

FinnPaes:
Implement Eventy in the import commands so hooking in custom data will not conflict with default Laravel wildcard events.

### DIFF
--- a/src/Commands/ImportBrands.php
+++ b/src/Commands/ImportBrands.php
@@ -5,8 +5,8 @@ namespace Rapidez\Statamic\Commands;
 use Statamic\Facades\Site;
 use Illuminate\Console\Command;
 use Rapidez\Statamic\Models\Brand;
+use TorMorten\Eventy\Facades\Eventy;
 use Rapidez\Statamic\Actions\StatamicEntryAction;
-use Illuminate\Support\Facades\Event;
 
 class ImportBrands extends Command
 {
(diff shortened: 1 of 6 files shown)


src/Commands/ImportBrands.php
```diff
@@ -40,7 +40,7 @@ public function handle(StatamicEntryAction $statamicEntryAction): int
                     array_merge([
                         'locale' => $site->handle(),
                         'site' => $site->handle(),
-                    ], ...Event::dispatch('rapidez-statamic:brand-entry-data', ['brand' => $brand]))
+                    ], ...[Eventy::filter('rapidez.statamic.brand.entry.data', $brand)])
```
indykoning, line 43: In this case destructuring wouldn't be needed anymore
```suggestion
                    ], Eventy::filter('rapidez.statamic.brand.entry.data', $brand))
```
Does everything still function the same as before? 
Can more data still be added to these fields without having to overwrite the original array of fields?

If not, it may be more beneficial to wrap the original `Event::dispatch` method in an array_filter to remove null and falsey values

FinnPaes: The functionality is not exactly the same as before. Because of this I took your advice to array_filter the data from the event. Did this in a new PR: #51 because almost everything in this PR would be irrelevant and the name of the commits wouldn't match the change.

FinnPaes:
I did some testing, Eventy works fine when the import commands are queued. I believe this will be the best option to give users the ability to extend data.

royduin:
Thanks! Looking good to me! Still needs a review from @indykoning

PR #51, opened 2024-01-11
Filter events from import commands

FinnPaes:
This fixes the possible NULL values if using Laravel wildcard events (https://github.com/rapidez/statamic/pull/50#discussion_r1447117258)


FinnPaes:
I will close this PR and open the other PR again, Eventy works fine when the command is queued.

PR #52, opened 2024-01-25
Moved responsive image observer to head stack

kevinmeijer97:


indykoning (approved):

config/rapidez/statamic.php
```diff
@@ -32,6 +32,8 @@
             'blueprint' => 'brand',
         ],
     ],
+    
+    'responsive_image_multiplier' => 150,
```
royduin, line 36: Maybe add a comment with some info?

src/RapidezStatamicServiceProvider.php
```diff
@@ -84,6 +85,7 @@ public function bootRoutes() : self
     public function bootViews() : self
     {
         $this->loadViewsFrom(__DIR__.'/../resources/views', 'rapidez-statamic');
+        $this->loadViewsFrom(__DIR__.'/../resources/views/vendor/responsive-images', 'responsive-images');
```
royduin, line 88: Why do we need this line?

kevinmeijer97: Had to specify it coming from `responsive-images` otherwise it wouldn't recognize it.
Had to actually commit the file also.. Did that in the commit below

PR #53, opened 2024-02-09
Implement optionalDeep add migration notice

indykoning:
BREAKING

Since optionalDeep will always return the optional class we explicitly need to ask it if a value is set
```diff
- @if($globals->cookie_notice->text)
+ @if($globals->cookie_notice->text->isset())
```

This also means that checks for specific values need to be updated
```diff
- @if($globals->cookie_notice->text === 'foo')
+ @if($globals->cookie_notice->text->get() === 'foo')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,33 @@ Eventy::addFilter('rapidez.statamic.brand.entry.data', fn($brand) => [
 
 When you create a form you could use `rapidez-statamic::emails.form` as HTML template which uses the [Laravel mail template](https://laravel.com/docs/master/mail#customizing-the-components) with all fields in a table, make sure you enable markdown!
 
+### Upgrading
+
+Since 0.4.0 we have started using optionalDeep for the $globals, and $content variables.
+This means some code may need to be upgraded. Here's a list of things you can expect might need to be changed:
+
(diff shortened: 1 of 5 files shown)


royduin:
Looking good! But your first example should still work?
```
@if($globals->cookie_notice->text)
```
If something doesn't exist it should return `null` right?

royduin:
Can't we somehow force `__toString()`?

Jade-GG:
> Can't we somehow force `__toString()`?

No, but you can always explicitly do it. The only way to theoretically have it work like this without a function is by abusing [simplexmlelement](https://www.php.net/manual/en/class.simplexmlelement.php), and having an expression turning it into a falsy xml element, that php will then parse into `false`. As you can imagine, this is not a proper solution and will not work for arbirtary objects like this.

Jade-GG:
It's been 15 years since [this original __toBoolean() proposal was rejected](https://externals.io/message/39323#39331) and it seems like a [currently open RFC](https://wiki.php.net/rfc/objects-can-be-falsifiable) also isn't going anywhere.

royduin:
Bummer, merged. Let's see how this turns out in projects.

PR #57, opened 2024-02-29
Use new runway path

Jade-GG:
They changed where this package is located.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "prefer-stable": true,
     "require": {
         "php": "^8.0|^8.1",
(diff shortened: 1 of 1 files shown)


Jade-GG:
> Should we also move on to version 6?

Almost certainly, yes, [the upgrade guide doesn't look too intimidating](https://runway.duncanmcclean.com/upgrade-guides/v5-x-to-v6-0). But we'll still have to put some time into making sure everything works properly after the update.

royduin:
Let's move on together with the upgrade as this old reference still works.

PR #58, opened 2024-03-08
Add @tags directive (and fix some styling)

Jade-GG:
Lets you use the `Statamic::tag()` function in a more convenient way, for example:

```blade
@tags([
    'types' => ['taxonomy:types' => ['color:is' => 'red']],
    'collection:pages',
])
```

Would give you two variables named `$types` and `$collectionPages` with the given statamic data.

See also [this PR](https://github.com/rapidez/blade-directives/pull/2). I don't know if this is really the right place either, as this would restrict this to Rapidez projects only while I seem to find the most benefits for this in non-rapidez Statamic projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,47 @@ Eventy::addFilter('rapidez.statamic.brand.entry.data', fn($brand) => [
 );
 ```
 
+### Blade directives
+
+#### @tags
+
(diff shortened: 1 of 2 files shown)


royduin:
Try a PR to `statamic/cms` if that's not accepted; `justbetter/statamic-helpers`

PR #59, opened 2024-03-25
Added overview and detail pages for brands

BobWez98:


resources/views/page_builder/brand_overview.blade.php
```diff
@@ -0,0 +1,52 @@
+@php
+    $brands = \Statamic\Facades\Entry::query()
+        ->where('collection', 'brands')
```
kevinmeijer97, line 3: This query is missing the site check.
`->where('site', \Statamic\Facades\Site::current()->handle())` should do the trick!

resources/views/brands/show.blade.php
```diff
@@ -0,0 +1,25 @@
...
+
+    @if ($title)
+        <h1 class="font-bold text-3xl mb-5">{{ $title }}</h1>
+    @endif
+    @if ($description)
+        <div class="prose prose-green max-w-none">
+            @content($description)
+        </div>
+    @endif
+    <x-rapidez::listing query="{ terms: { 'manufacturer.keyword': ['{{ $title }}'] } }"/>
```
royduin, line 22: This should listen to https://github.com/rapidez/statamic/blob/master/config/rapidez/statamic.php#L46

resources/views/brands/show.blade.php
```diff
@@ -0,0 +1,25 @@
+@extends('rapidez::layouts.app')
+
+@section('title', $meta_title->value() ?: $title)
+@section('description', $meta_description)
```
royduin, line 4: Are those fields included in the brand blueprint?

resources/content/collections/brands.yaml
```diff
@@ -1,8 +1,14 @@
-title: 'Brand content'
+title: Brands
```
royduin, line 1: We called it "Brand content" just like the categories and products as there is also a Raleway resource. Otherwise we've 2x brands.

resources/content/collections/brands.yaml
```diff
@@ -1,8 +1,14 @@
...
 propagate: true
+template: rapidez-statamic::brands/show
+layout: layout
 revisions: false
-slugs: false
+route: '/brands/{slug}'
```
royduin, line 6: Is it possible to make this configurable? Should this be de default? Otherwise we should mention this in the readme.

resources/views/page_builder/brand_overview.blade.php
```diff
@@ -0,0 +1,53 @@
...
+                        <a href="{{ $brand->url() }}">{{ $brand->title }}</a>
+                    </li>
+                @endforeach
+            </ul>
+        </div>
+    @endif
+    <div class="flex flex-col">
+        @if($page->title)
+            <h1>{{ $page->title }}</h1>
+        @endif
```
royduin, line 29: Don't think we should include this. If a brands overview page is build like with content before this component; most likely the page title is included there.

resources/views/brands/show.blade.php
BobWez98: Now they are.

resources/content/collections/brands.yaml
kevinmeijer97: I would leave this default to brands, if someone wants it changed they can do so in Statamic.
Updating the readme with the new changes would be a nice anyways!

royduin: We could also leave it disabled; if someone wants it they can enable it with info in the readme. When you install this package; should brand pages work automatically? Can imagine not everyone want/needs this.

kevinmeijer97: True, in that case we could also listen to the brand enabled option in the config when publishing the yaml files?

BobWez98: No we couldn't because then we need to either programmaticly edit the yaml file or add a custom route for it. I would say just leave it out for now.

BobWez98: And add a description in the readme of how to enable it.

resources/views/brands/show.blade.php
```diff
@@ -0,0 +1,27 @@
...
+@push('head')
+    @foreach(Statamic::tag('alternates')->params(compact('page')) as $lang => $url)
+        <link rel="alternate" hreflang="{{ $lang }}" href="{{ $url }}" />
+    @endforeach
+@endpush
```
royduin, line 9: Does this work? With multilingual brand pages?

resources/views/brands/show.blade.php
```diff
@@ -0,0 +1,27 @@
...
+    @includeWhen(!$is_homepage, 'rapidez-statamic::breadcrumbs')
+
+    @if ($title)
+        <h1 class="font-bold text-3xl mb-5">{{ $title }}</h1>
+    @endif
+    @if ($description)
+        <div class="prose prose-green max-w-none">
+            @content($description)
+        </div>
+    @endif
```
royduin, line 23: This doesn't exist in the blueprint? Maybe just move the page builder include here?

resources/content/collections/brands.yaml
royduin: 👌🏻

resources/views/page_builder/brand_overview.blade.php
```diff
@@ -0,0 +1,50 @@
...
+            <span>@lang('Brands')</span>
+            <ul class="flex flex-col">
+                @foreach($brands as $brand)
+                    <li>
+                        <a href="{{ $brand->url() }}">{{ $brand->title }}</a>
+                    </li>
+                @endforeach
+            </ul>
+        </div>
+    @endif
```
royduin, line 25: Maybe a bit weird to have 2 brand listings?

resources/views/page_builder/brand_overview.blade.php
```diff
@@ -0,0 +1,50 @@
...
+
+        <div class="flex flex-col">
+            @foreach($grouped as $letter => $group)
+                <div class="flex w-full">
+                    <span id="{{ $letter }}" class="text-xl">
+                        {{ $letter }}
+                    </span>
+                    <ul class="flex row-wrap">
+                        @foreach($group as $brand)
+                            <li>{{ $brand->title }}</li>
```
royduin, line 43: No brand image?

resources/views/brands/show.blade.php
BobWez98: I added it to the blueprint. I'd say we only want a description above the product listing. If we want to add other content you'd want it below the product grid.

resources/blueprints/collections/brands/brand.yaml
```diff
@@ -24,6 +24,51 @@ sections:
           instructions_position: above
           visibility: visible
           localizable: true
+      - handle: description
```
royduin, line 27: Better use the fieldset: https://github.com/rapidez/statamic/blob/master/resources/fieldsets/content.yaml, this way the configuration is in 1 place.

resources/fieldsets/page_builder.yaml
```diff
@@ -22,6 +22,10 @@ fields:
           fields:
             -
               import: product_slider
+        brand_overview:
```
royduin, line 25: Mabye rename it to `brandlist`?

resources/blueprints/collections/brands/brand.yaml
BobWez98: That's not possible as the pagebuilder is in this. The content fieldset and the pagebuilder have the same handle so they would overwrite eachother.

resources/fieldsets/brand_list.yaml
```diff
@@ -0,0 +1,12 @@
+title: 'Brand overview'
```
royduin, line 1: Brandlist? Also as filename? To keep things consistent.

resources/blueprints/collections/brands/brand.yaml
royduin: If I remember correctly; it's possible to give it another handle

PR #60, opened 2024-03-29
Download images from amasty brands if they're present

BobWez98:


src/Commands/ImportBrands.php
```diff
@@ -2,34 +2,53 @@
...
             foreach (Brand::get() as $brand) {
+                $extraData = [];
+
+                if (Schema::hasTable('amasty_amshopby_option_setting')) {
+                    $amastyData = DB::table('amasty_amshopby_option_setting')->where('value', '=', $brand->option_id)->first();
+                    $extraData['meta_title'] = $amastyData->meta_title;
+                    $extraData['meta_description'] = $amastyData->meta_description;
+
+                    if ($amastyData->image) {
+                        $file = file_get_contents(config('rapidez.media_url').'/amasty/shopby/option_images/'.$amastyData->image, false, stream_context_create(['ssl' => ['verify_peer' => false, 'verify_peer_name' => false]]));
```
indykoning, line 46: This will break when the image cannot be downloaded for any reason, like it being missing, or the host not reachable.

Is this intended? if not, take a look at how we download the image in the image resizer module https://github.com/rapidez/image-resizer/blob/023cf6a6d4111ab523d3f5659cf2343e2ff72301/src/Controllers/ImageController.php#L131

indykoning (approved):

PR #65, opened 2024-05-14
Get site based on Magento store ID

kevinmeijer97:


src/RapidezStatamicServiceProvider.php
```diff
@@ -218,4 +218,13 @@ public function currentSiteIsEnabled(): bool
     {
         return !config('statamic.sites.sites.' . Site::current()->handle() . '.attributes.disabled', false);
     }
+
+    public function getSiteHandleByStoreId(): string
+    {
+        $site = Site::all()
+            ->filter(fn($_site) => ($_site?->attributes()['magento_store_id'] ?? null) === config('rapidez.store'))
```
Jade-GG, line 225: Should we have a fallback where it checks the store code if this isn't set, as to not break backwards compatibility?

kevinmeijer97: Sure, we fallback to the store_code if no site's been found.

Jade-GG (approved):

PR #67, opened 2024-07-30
Added Statamic 5 support + install command

kevinmeijer97:


README.md
```diff
@@ -21,6 +21,28 @@ You need to have `statamic/cms` installed in your Rapidez installation. Follow t
...
 
+## Install command
+
+The install command will help you set up all the necessary settings.
+
+```
+php artisan rapidez-statamic:install
+```
+
+Please make sure to use the following settings when setting up the eloquent driver:
```
BobWez98, line 32: What are the settings? Maybe add a stub with a pre configured config file?

README.md
```diff
@@ -57,6 +79,8 @@ If you'd like to use the homepage from Statamic instead of the CMS page from Mag
 
 ### Publish Collections, Blueprints and Fieldsets
 
+If you ran the install command these will already have been published.
```
BobWez98, line 82: If you have run the ....

README.md
```diff
@@ -70,21 +94,18 @@ php artisan vendor:publish --provider="Rapidez\Statamic\RapidezStatamicServicePr
...
+It is important to add the Magento store ID for every site in the attributes section within `resources/sites.yaml` and use the store code as key. The current site will be determined based on the `MAGE_RUN_CODE`. By default Statamic uses the url for this; that's still the fallback. If you need to generate some urls with a multisite it's a best practice to specify the `url` per site from env variables. See the [Statamic multisite docs](https://statamic.dev/multi-site#url). Optionally you could set the `group` within the `attributes` if you'd like to group sites to filter the alternate hreflang link tags. You could also set the `disabled` within the `attributes` section to true if you want to exclude this site from being altered with Statamic data.
+
+```yaml
+default:
+  name: '{{ config:app:name }}'
+  url: /
+  locale: nl_NL
```
BobWez98, line 103: maybe add a second key for lang: nl_NL too?

src/Commands/InstallCommand.php
```diff
@@ -0,0 +1,54 @@
...
+    public function handle(): void
+    {
+        $this->call('vendor:publish', [
+            '--tag' => 'runway-config',
+        ]);
+
+        $this->call('vendor:publish', [
+            '--provider' => 'Statamic\Eloquent\ServiceProvider',
+            '--tag' => 'statamic-eloquent-config',
+        ]);
```
BobWez98, line 22: Maybe ask right away which config the user wants published; https://laravel.com/docs/11.x/artisan#multiple-choice-questions, if the publishing is mandatory, nevermind this comment.

src/Commands/InstallCommand.php
```diff
@@ -0,0 +1,54 @@
...
+
+        $this->confirm('Did you set the table_prefix in config/statamic/eloquent-driver.php to "statamic_"?');
```
BobWez98, line 24: Could we do this for the user?

src/Commands/InstallCommand.php
BobWez98: Or maybe also include the `Collections, Blueprints and Fieldsets` and `Views` in this? Just add the 4 options and "all" Or we could try making a multiple select like the default `vendor:publish` command.
Scherm­afbeelding 2024-07-30 om 14 49 35

src/Models/Brand.php
```diff
@@ -37,7 +37,7 @@ protected static function booting()
                 })
                 ->leftJoinSub($renamedStore, 'store_value', function ($join) {
                     $join->on('store_value.sub_option_id', '=', 'eav_attribute_option.option_id')
-                         ->where('store_value.store_id', (Statamic::isCpRoute() ? Site::selected()->attributes['magento_store_id'] : Site::current()->attributes['magento_store_id']));
+                         ->where('store_value.store_id', (Statamic::isCpRoute() ? (Site::selected()->attributes['magento_store_id'] ?? '1') : (Site::current()->attributes['magento_store_id'] ?? '1')));
```
BobWez98, line 40: Why the `?? 1` , maybe we also need a `validate` / healthcheck command that checks if the mandatory config values are set.

BobWez98 (changes requested):

composer.json
```diff
@@ -30,11 +30,13 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "php": "^8.0|^8.1",
-        "doublethreedigital/runway": "^5.0",
-        "rapidez/blade-directives": "^0.4",
-        "spatie/statamic-responsive-images": "^4.0",
-        "statamic/cms": "^4.0",
+        "php": "^8.1|^8.2",
+        "justbetter/statamic-eloquent-driver-globalset-migration-generator": "^0.1.0",
```
royduin, line 34: Is this still required? Some PR's related to this where opened / merged in Statamic / the Eloquent driver.

src/Commands/InstallCommand.php
```diff
@@ -0,0 +1,54 @@
...
+
+        $this->call('statamic:install:eloquent-driver');
+
+        $this->info('If you would like to store users in the database please make sure to follow this guide: https://statamic.dev/tips/storing-users-in-a-database#from-a-fresh-statamic-project');
```
royduin, line 28: Pro / cons? What do we prefer and why?

src/Commands/InstallCommand.php
```diff
@@ -0,0 +1,54 @@
...
+
+        if ($this->confirm('Would you like to configure a multisite?')) {
```
royduin, line 30: I remember a something a while ago when a site needed to be multisite to work correct, is this still te case? Do we really support both?

src/Commands/InstallCommand.php
```diff
@@ -0,0 +1,54 @@
...
+            $this->call('statamic:multisite');
+        }
+
+        if ($this->confirm('Would you like to publish Collections, Blueprints and Fieldsets?')) {
```
royduin, line 34: If I get a question like this I always think; should I? Why and in what cases is this useful? Can we clarify it or if it's a good practice just do it?

src/Commands/InstallCommand.php
```diff
@@ -0,0 +1,54 @@
...
+            $this->call('vendor:publish', [
+                '--provider' => 'Rapidez\Statamic\RapidezStatamicServiceProvider',
+                '--tag' => 'rapidez-statamic-content',
+            ]);
+        }
+
+        if ($this->confirm('Would you like views?')) {
```
royduin, line 41: Same as above;
> If I get a question like this I always think; should I? Why and in what cases is this useful? Can we clarify it or if it's a good practice just do it?

royduin:
What we're missing is the user stuff we need to copy over from `laravel/laravel` like the user model and the migrations. Or should we bring that back into `rapidez/rapidez`? Currently it's removed: https://github.com/rapidez/rapidez/blob/master/.github/workflows/copy-laravel-source.yml#L45:L46

royduin:
Maybe some extra info in the readme on the `$globals`?

src/Commands/InstallCommand.php
indykoning: We can definitely verify this for the user. At least checking if `config('statamic.eloquent-driver.table_prefix')` is not empty

src/Models/Brand.php
indykoning: You can check out https://github.com/rapidez/paynl/pull/21/files#diff-372b727c0d9abc808461b82d701bcd8df3609a9396e050bd5f21ad6515d05f05 to see how to make the health check.

You can add as many messages as you like and determine wether is means a health check fail when you want.

README.md
kevinmeijer97: Because of changing config files from the eloquent driver i would rather document what to select for eloquent here rather than having an outdated config file.

README.md
kevinmeijer97: Noted, will look into this later!

src/Commands/InstallCommand.php
kevinmeijer97: Made the publishing semi-mandatory by setting the default to yes, still want to give the choice to the user but make it known that default is to publish.

src/Models/Brand.php
kevinmeijer97: Noted, will look into this later!

composer.json
```diff
@@ -30,11 +30,13 @@
...
+        "rapidez/blade-directives": "^0.6",
+        "spatie/statamic-responsive-images": "^5.0",
```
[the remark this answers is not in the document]
kevinmeijer97: Noted, will look into this later!

src/Commands/InstallCommand.php
kevinmeijer97: Since we require the eloquent driver we should also keep the users in the database as well.
Noted this to fix later!

src/Commands/InstallCommand.php
kevinmeijer97: The package needs to publish so i set the default to yes for these questions.

src/Commands/InstallCommand.php
kevinmeijer97: See above!

composer.json
kevinmeijer97: Noted, will look into this later!

src/Commands/InstallCommand.php
kevinmeijer97: Multisite would make it that the user has to buy the PRO version of Statamic. Noted and will look into using a non multisite setup later!

PR #71, opened 2024-09-03
Fix brands uri generation

indykoning:
Brands would not generate uris for brands before, this should let it generate


royduin:
This is mentioned in the readme and I think this should stay optional: https://github.com/rapidez/statamic#brand-pages

BobWez98 (approved):

BobWez98 (commented):
Yes, we specificly left this out to let it be optional as @royduin said.

indykoning:
As mentioned in the conversations, this has been left out on purpose.

PR #74, opened 2024-09-16
Add desktop/mobile navigation

JimmyHoenderdaal:


resources/views/components/nav-layer.blade.php
```diff
@@ -0,0 +1,57 @@
...
+                <li>
+                    <a href="{{ $parentUrl }}" class="normal flex items-center justify-between p-5 font-semibold">
+                        @lang('Go to :item', ['item' => strtolower($title)])
+                    </a>
+                </li>
+            @endif
+            @foreach ($children ?: [] as $child)
+                @php
+                    $url = $child['url'];
+                    if (isset($child['linked_category'])) {
```
BobWez98, line 27: I don't think `linked_category` is a default field, can you add the blueprint?

resources/views/components/nav-layer.blade.php
```diff
@@ -0,0 +1,57 @@
+@props(['id', 'children', 'title' => __('Menu'), 'hasParent' => false, 'tag' => 'form', 'parentUrl' => ''])
+@slots(['headerbutton'])
+
+<x-rapidez::slideover.mobile
+    :title="(string) $title"
+    :$id
+    :$hasParent
+    :$tag
+>
+    <x-slot:headerbutton>
```
BobWez98, line 10: i think header-button would be cleaner

resources/views/navigation/nav.blade.php
```diff
@@ -0,0 +1,54 @@
...
+    />
+</nav>
+@php
+    $baseUrl = \Statamic\Facades\Site::current()->absoluteUrl();
+@endphp
+<nav class="relative border-y max-lg:hidden">
+    <ul class="flex items-center justify-center gap-8 font-semibold text-neutral [&>:not(:hover)]:hover:text-inactive text-sm">
+        @foreach (Statamic::tag('nav:main')->fetch() as $item)
+            @php
+                $level1Url = isset($item['linked_category']) && ($item['linked_category']?->value()['url_path'] ?? false) ? $baseUrl . '/' . $item['linked_category']->value()['url_path'] : $item['url'] ?? '';
```
BobWez98, line 15: I see this check done multiple times, maybe we can add a helper function for it or something? Maybe like these: [https://github.com/rapidez/core/blob/master/src/helpers.php](https://github.com/rapidez/core/blob/master/src/helpers.php) ? Or even nicer maybe add a facade `RapidezStatamic::getNavUrl($item)` like we have for [the config](https://github.com/rapidez/core/blob/master/src/Rapidez.php#L42)

JimmyHoenderdaal:
image

image

resources/views/navigation/nav.blade.php
BobWez98: Also as a addition i don't really like the level1, level2, level3 etc, can't you just call each iteration `$item` ? instead of; 
```
@foreach ($item['children'] as $level2)
```
do
```
@foreach ($item['children'] as $item)
```

BobWez98 (changes requested):

resources/views/components/nav-layer.blade.php
JimmyHoenderdaal: that would have to to be changed in rapidez/core for the slideover

BobWez98: Nevermind then

resources/views/navigation/header-button.blade.php
```diff
@@ -0,0 +1,2 @@
+{{-- This is shown on the first slideover of the mobile menu, you can leave this empty if you don't want it. By default this is a search button --}}
+<x-button.search />
```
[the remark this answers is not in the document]
JimmyHoenderdaal: yes, I'm not sure what you mean by this

resources/views/components/nav-layer.blade.php
```diff
@@ -0,0 +1,57 @@
+@props(['id', 'children', 'title' => __('Menu'), 'hasParent' => false, 'tag' => 'form', 'parentUrl' => ''])
+@slots(['headerbutton'])
+
+<x-rapidez::slideover.mobile
+    :title="(string) $title"
```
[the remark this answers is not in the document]
JimmyHoenderdaal: otherwise I get this error
image

resources/views/components/nav-layer.blade.php
```diff
@@ -0,0 +1,57 @@
...
+                    @if ($child['title'] ?? '')
+                        <a href="{{ $url }}" class="flex items-center justify-between p-5 font-semibold">
+                            {{ $child['title'] }}
+                            @if ($child['children'])
+                                <x-heroicon-o-chevron-right class="size-4" />
+                            @endif
+                        </a>
+                    @endif
+                    @if ($child['children'])
+                        @php($childId = uniqid(Str::snake("{$child['title']}" ?? '')))
```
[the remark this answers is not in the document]
JimmyHoenderdaal: this is using uniqid but it adds snake case to the title for whenever there are spaces in the title
image

I want to use the title here since it makes it more clear what id's and for's are linked to each other in your browser.
Could of course remove that part and only have uniqid but that would make the id a bit abstract and thus hard to debug if needed

resources/views/navigation/header-button.blade.php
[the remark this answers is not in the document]
BobWez98: This component does not have a namespace e.g. `<x-rapidez::button.search />` and thus this will only work when you add this in your project. This component either needs to be added to this package, come from a package in it's dependencies or be removed.

BobWez98 (changes requested):

resources/views/navigation/header-button.blade.php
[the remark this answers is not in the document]
JimmyHoenderdaal: I see thanks for the clarification, will leave this blank so you can overwrite it in the project itself :)

resources/views/components/nav-layer.blade.php
BobWez98: It actually is **not**. on all our projects this is a field on the navigation blueprint, not on the category itself.

BobWez98:
This is added in #79 .

PR #77, opened 2024-09-30
Implement performance optimizations

indykoning:
This implements an array cache to some often used functions.
Quick breakdown:
`FindByUrl` and by extension `findByMageRunCode` is getting hit `313` times.
On single store (or simply without the MAGE_RUN_CODE) this would result in 626 creations and lookups in a collection.
if the code is missing the `findByMageRunCode` cache is never created, thus the falsey check saves us on 313 lookups in case that header is missing. Otherwise we can assume the site will be found.

the casting of the large array of globals to object, and passing it to optionalDeep seems relatively heavy.
and it is executed `284` times (per view it is executed)
We cannot store it in Redis cache without breaking the object, thus array store is the best option here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "justbetter/statamic-eloquent-driver-globalset-migration-generator": "^0.1.0",
         "rapidez/blade-directives": "^0.6",
         "justbetter/statamic-glide-directive": "^2.1",
+        "spatie/once": "*",
         "statamic-rad-pack/runway": "^7.6",
         "statamic/cms": "^5.0",
         "statamic/eloquent-driver": "^4.9",
--- a/src/Extend/SitesLinkedToMagentoStores.php
+++ b/src/Extend/SitesLinkedToMagentoStores.php
@@ -2,7 +2,7 @@
 
 namespace Rapidez\Statamic\Extend;
 
-use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Collection;
 use Statamic\Sites\Sites;
 
 class SitesLinkedToMagentoStores extends Sites
@@ -19,9 +19,9 @@ class SitesLinkedToMagentoStores extends Sites
 
     public function findByMageRunCode($code)
     {
-        if (!$code) {
+        if (!$code || !($this->sites instanceof Collection)) {
             return null;
         }
-        return Cache::store('array')->rememberForever('rapidez.statamic.findByMageRunCode-'.$code, fn() => collect($this->sites)->get($code));
+        return $this->sites->get($code));
     }
 }
--- a/src/Extend/SitesLinkedToMagentoStores.php
+++ b/src/Extend/SitesLinkedToMagentoStores.php
@@ -9,11 +9,11 @@ class SitesLinkedToMagentoStores extends Sites
 {
     public function findByUrl($url)
     {
-        if ($site = $this->findByMageRunCode(request()->server('MAGE_RUN_CODE'))) {
+        if ($site = once(fn() => $this->findByMageRunCode(request()->server('MAGE_RUN_CODE')))) {
             return $site;
         }
 
-        return Cache::store('array')->rememberForever('rapidez.statamic.findByUrl-'.$url, fn() => parent::findByUrl($url));
+        return once(fn() => parent::findByUrl($url));
     }
 
 
--- a/src/Extend/SitesLinkedToMagentoStores.php
+++ b/src/Extend/SitesLinkedToMagentoStores.php
@@ -2,6 +2,7 @@
 
 namespace Rapidez\Statamic\Extend;
 
+use Illuminate\Support\Facades\Cache;
 use Statamic\Sites\Sites;
 
 class SitesLinkedToMagentoStores extends Sites
@@ -12,12 +13,15 @@ class SitesLinkedToMagentoStores extends Sites
             return $site;
         }
 
-        return parent::findByUrl($url);
+        return Cache::store('array')->rememberForever('rapidez.statamic.findByUrl-'.$url, fn() => parent::findByUrl($url));
     }
 
 
     public function findByMageRunCode($code)
     {
-        return collect($this->sites)->get($code);
+        if (!$code) {
+            return null;
+        }
+        return Cache::store('array')->rememberForever('rapidez.statamic.findByMageRunCode-'.$code, fn() => collect($this->sites)->get($code));
     }
 }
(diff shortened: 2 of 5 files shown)


BobWez98 (dismissed):

composer.json
```diff
@@ -34,6 +34,7 @@
         "justbetter/statamic-eloquent-driver-globalset-migration-generator": "^0.1.0",
         "rapidez/blade-directives": "^0.6",
         "justbetter/statamic-glide-directive": "^2.1",
+        "spatie/once": "*",
```
royduin, line 37: This is added in Laravel 11: https://laravel.com/docs/11.x/releases#the-once-function, active support for Laravel 10 ended, see: https://endoflife.date/laravel we could move on but maybe we need to require `illuminate/support:^11.0`

src/Http/ViewComposers/StatamicGlobalDataComposer.php
```diff
@@ -31,7 +31,7 @@ public function getGlobals()
     public function compose(View $view) : View
     {
         if (!$this->globals) {
-            $this->globals = optionalDeep((object)$this->getGlobals());
+            $this->globals = Cache::driver('array')->rememberForever('rapidez.statamic.globals', fn() => optionalDeep((object)$this->getGlobals()));
```
royduin, line 34: This is already "cached" by the singleton right?

src/Models/Category.php
```diff
@@ -15,6 +15,6 @@ class Category extends Model
 
     public function getTable()
     {
-        return 'catalog_category_flat_store_'.(Statamic::isCpRoute() ? (Site::selected()->attributes['magento_store_id'] ?? '1') : (Site::current()->attributes['magento_store_id'] ?? '1'));
+        return 'catalog_category_flat_store_'.once(fn() => (Statamic::isCpRoute() ? (Site::selected()->attributes['magento_store_id'] ?? '1') : (Site::current()->attributes['magento_store_id'] ?? '1')));
```
royduin, line 18: Is this going as expected when switching between de frontend and the backend? And switching sites from the backend?

src/Extend/SitesLinkedToMagentoStores.php
```diff
@@ -12,12 +13,15 @@ public function findByUrl($url)
...
     }
 
 
     public function findByMageRunCode($code)
     {
-        return collect($this->sites)->get($code);
+        if (!$code) {
+            return null;
+        }
+        return Cache::store('array')->rememberForever('rapidez.statamic.findByMageRunCode-'.$code, fn() => collect($this->sites)->get($code));
```
royduin, line 25: Sites should be a collection already, so `$this->sites->ge($code)` should do the same, as `Sites` is a singleton I think any caching here doesn't have any effect.

src/Extend/SitesLinkedToMagentoStores.php
```diff
@@ -12,12 +13,15 @@ public function findByUrl($url)
             return $site;
         }
 
-        return parent::findByUrl($url);
+        return Cache::store('array')->rememberForever('rapidez.statamic.findByUrl-'.$url, fn() => parent::findByUrl($url));
```
royduin, line 16: So this is really faster? With or without; both will get data from an array eventually.

composer.json
indykoning: We don't necessarily need to require L11+ for this functionality.
If we require it like this it'll install `spatie/once` for <11 and `illuminate/support` will overwrite it on >=11
https://packagist.org/packages/illuminate/support#:~:text=collect%3A%20%3C5.5.33-,replaces,-spatie/once%3A%20*

src/Http/ViewComposers/StatamicGlobalDataComposer.php
indykoning: I expected that as well, during profiling it did not appear to be a singleton. Instead calling optionalDeep for every view.

src/Models/Category.php
indykoning: It should, since once only memoizes the data per response. Switching store would use another request, thus no cache the first time.
The only case where this could be an issue is while switching store within the same request which doesn't happen.
(We do switch the store within the same request in rapidez:index but that does not impact this model)

src/Extend/SitesLinkedToMagentoStores.php
indykoning: With the amount of times this was executing the constant casting to array, and from array to collection the caching did have an effect.

However, if we make sure it's a collection (which it should be) we can skip the `collection -> array -> collection`
I've updated the code

src/Http/ViewComposers/StatamicGlobalDataComposer.php
royduin: Mm, that's weird. With my profiling earlier it worked great: https://github.com/rapidez/statamic/pull/73/files + https://github.com/rapidez/statamic/blob/master/src/RapidezStatamicServiceProvider.php#L158

BobWez98:
@indykoning what's the status?

src/Http/ViewComposers/StatamicGlobalDataComposer.php
indykoning: I agree it shouldn't have impact, for now i will remove this.
We may get back to it in the future, but it shouldn't be necessary

BobWez98 (approved):

kevinmeijer97 (approved):

src/Models/Category.php
royduin, line 18:
```suggestion
        return 'catalog_category_flat_store_'.once(fn() => (Statamic::isCpRoute()
            ? (Site::selected()->attributes['magento_store_id'] ?? '1')
            : (Site::current()->attributes['magento_store_id'] ?? '1')));
```

src/Models/Product.php
```diff
@@ -24,6 +24,6 @@ protected static function booting()
 
     public function getTable()
     {
-        return 'catalog_product_flat_' . (Statamic::isCpRoute() ? (Site::selected()->attributes['magento_store_id'] ?? '1') : Site::current()->attributes['magento_store_id'] ?? '1');
+        return 'catalog_product_flat_' . once(fn() => (Statamic::isCpRoute() ? (Site::selected()->attributes['magento_store_id'] ?? '1') : Site::current()->attributes['magento_store_id'] ?? '1'));
```
royduin, line 27:
```suggestion
        return 'catalog_product_flat_' . once(fn() => (Statamic::isCpRoute()
            ? (Site::selected()->attributes['magento_store_id'] ?? '1')
            : Site::current()->attributes['magento_store_id'] ?? '1'));
```

PR #78, opened 2024-10-04
Added sitemap for collections & taxonomies

kevinmeijer97:


src/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,148 @@
+<?php
```
BobWez98, line 1: Will this command always finish? Or run into a timeout maybe? We have a project where there's 23K entries (divided over multiple shops) of a collection that should be in the sitemap. Maybe create Jobs for each sitemap so we can schedule this in a queue and do it in smaller operations? Also would e nice to be able to schedule this once or twice per day.

src/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $taxonomies = TaxonomyFacade::all()
+            ->filter(function ($taxonomy) use ($site) {
+                $terms = $taxonomy->queryTerms()
+                    ->where('site', $site->handle())
+                    ->get();
+
+                return $terms
+                    ->filter(function ($term) {
+                        return view()->exists($term->template());
```
BobWez98, line 56: can't you check for a template in the query? It's a bit more efficient, also don't have to check the is not empty then

src/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $sitemap .= '</sitemapindex>';
+        file_put_contents(public_path('sitemap_statamic_' . $site->handle() . '.xml'), $sitemap);
+    }
+
+    protected function generateCollectionSitemap(Site $site, Collection $collection) : void
+    {
+        $sitemap = '<?xml version="1.0" encoding="UTF-8"?>';
+        $sitemap .= '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml">';
+        $entries = $collection->queryEntries()->where('site', $site->handle())->whereStatus('published')->get();
```
BobWez98, line 74: We can have collections that have no route, like Category content, but have no url, how is this handled? I see no check on the `->url()`

README.md
```diff
@@ -236,6 +236,20 @@ For example; If you created a global with the handle `header` and added a field
...
+It is possible to create a sitemap per store, each sitemap is a index containing all collections & taxonomies that contain their own views.
+The XML files will be generated as:
+```shell
+sitemap_statamic_{site_handle}
+sitemap_statamic_collection_{site_handle}_{collection_handle}
+sitemap_statamic_taxonomy_{site_handle}_{taxonomy_handle}
+```
+You can generate the sitemaps by running:
+```shell
+rapidez:statamic:generate:sitemap
```
BobWez98, line 250: add `php artisan`

BobWez98 (changes requested):

README.md
```diff
@@ -236,6 +236,20 @@ For example; If you created a global with the handle `header` and added a field
 
 When you create a form you could use `rapidez-statamic::emails.form` as HTML template which uses the [Laravel mail template](https://laravel.com/docs/master/mail#customizing-the-components) with all fields in a table, make sure you enable markdown!
 
+### Sitemap
+
+It is possible to create a sitemap per store, each sitemap is a index containing all collections & taxonomies that contain their own views.
+The XML files will be generated as:
```
royduin, line 242:
```suggestion
It is possible to create a sitemap per store, each sitemap is an index containing all collections and taxonomies that do have a views. The XML files will be generated as:
```

README.md
```diff
@@ -236,6 +236,20 @@ For example; If you created a global with the handle `header` and added a field
...
+```shell
+sitemap_statamic_{site_handle}
+sitemap_statamic_collection_{site_handle}_{collection_handle}
+sitemap_statamic_taxonomy_{site_handle}_{taxonomy_handle}
```
royduin, line 246:
```suggestion
/sitemap_statamic_{site_handle}.xml
/sitemap_statamic_collection_{site_handle}_{collection_handle}.xml
/sitemap_statamic_taxonomy_{site_handle}_{taxonomy_handle}.xml
```

src/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,148 @@
...
+use Statamic\Taxonomies\Taxonomy;
+use Statamic\Taxonomies\LocalizedTerm;
+
+class GenerateSitemap extends Command
+{
+    protected $signature = 'rapidez:statamic:generate:sitemap';
+
+    protected $description = 'Generate a sitemap based on Statamic collections & taxonomies';
+
+
```
royduin, line 22:
```suggestion

```

src/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,148 @@
...
+            ->filter(function ($term) {
+                return view()->exists($term->template());
+            });
+    }
+
+    protected function addSitemap($url = null) : string
+    {
+        $sitemap = '<sitemap>';
+
+        if ($url ?? false) {
```
royduin, line 123: `$url` is a method param so always there

```suggestion
        if ($url) {
```

royduin (changes requested):
Is checking the view the right way? Can't we check if the collection/taxonomy has a route?

src/Commands/GenerateSitemap.php
kevinmeijer97: The collections that are queried are already filtered by collections that have urls, so collections like the category content won't come here

kevinmeijer97:
> Is checking the view the right way? Can't we check if the collection/taxonomy has a route?

Sadly we cant check routes on taxonomies the same way we can check routes on collections.
For the taxonomies this is also the way Statamic seems to do it for their sitemap in the SEO addon.

src/Commands/GenerateSitemap.php
kevinmeijer97: See https://github.com/rapidez/statamic/pull/78#issuecomment-2399337106
Taxonomies are limited in the way we can check for urls, Statamic seems to be taking this approach themselves as well.

BobWez98: That's not what i'm saying, you can skip the whole `->filter()` if you doe a `->where` on the template in the query

kevinmeijer97: No thats not possible, the template is not something we can query on. You can see why here https://github.com/statamic/cms/blob/5.x/src/Taxonomies/LocalizedTerm.php#L381

kevinmeijer97: Was just able to test it, unfortunately an existing `uri` column doesn't guarantee an existing view. It seems like as long as a term has a slug, the `uri` field will be filled.

This is probably also why Statamic does it this way themselves in their SEO addon.

kevinmeijer97: Either way, still added the uri check in the query. Since all terms that do have views also do have uri's.

royduin (changes requested):
When generated they aren't ignored by the `.gitignore` by default I think?

kevinmeijer97:
> When generated they aren't ignored by the `.gitignore` by default I think?

Correct, i added a line in the readme for this now!

src/RapidezStatamicServiceProvider.php
```diff
@@ -2,7 +2,6 @@
```
BobWez98, line 2: De we maye need to schedule the jobs?

src/Jobs/GenerateTermSitemapJob.php
```diff
@@ -0,0 +1,63 @@
...
+        protected Taxonomy $taxonomy,
+    ) {
+        $this->sitemapGenerator = new (GenerateSitemapAction::class);
+    }
+
+    public function handle(): void
+    {
+        $sitemap = $this->sitemapGenerator->createSitemap($this->generateTermSitemap());
+
+        file_put_contents(public_path('sitemap_statamic_taxonomy_' . $this->site->handle() . '_' . $this->taxonomy->handle() . '.xml'), $sitemap);
```
indykoning, line 33: It's safer to use the [Laravel Storage facade](https://laravel.com/docs/11.x/filesystem#storing-files) for this

src/Jobs/GenerateCollectionSitemapJob.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    protected GenerateSitemapAction $sitemapGenerator;
+
+    public function __construct(
+        protected Site $site,
+        protected Collection $collection,
+    ) {
+        $this->sitemapGenerator = new (GenerateSitemapAction::class);
```
royduin, line 26: Can't we inject this?

composer.json
```diff
@@ -31,9 +31,10 @@
     "prefer-stable": true,
     "require": {
         "php": "^8.1|^8.2",
+        "justbetter/statamic-glide-directive": "^2.1",
         "rapidez/blade-directives": "^0.6",
         "rapidez/core": "^2.13",
-        "justbetter/statamic-glide-directive": "^2.1",
+        "rapidez/sitemap": "dev-master",
```
royduin, line 37: I've tagged a version so this dependency can be updated

src/RapidezStatamicServiceProvider.php
```diff
@@ -213,6 +218,41 @@ public function bootUtilities() : static
         return $this;
     }
 
+    public function bootSitemaps(): static
+    {
+        Schedule::command('rapidez:statamic:generate:sitemap')->twiceDaily(0, 12);
```
royduin, line 223: Not sure if we want to force this from the package, maye put it somewhere in the readme? Or provide a command from `rapidez/sitemap` where this package hooks into. So you'll just get 1 command which you can schedule yourself which generates all sitemaps.

README.md
```diff
@@ -54,6 +54,21 @@ We recommend migrating the following options to the database when setting up the
 
 After you're done running the install command make sure to check our configuration guide written [here](https://docs.rapidez.io/3.x/packages/statamic.html).
 
+### Sitemap
+
+It is possible to create a sitemap per store, each sitemap is an index containing all collections and taxonomies that do have views. The XML files will be generated as:
```
BobWez98, line 59: Is this still true?

config/rapidez/statamic.php
```diff
@@ -94,4 +94,9 @@
     'disabled_sites' => [
         // '{site_handle}'
     ],
+
+    'sitemap' => [
+        'prefix' => 'statamic_sitemap_',
```
BobWez98, line 99: Maybe add an env value for this? That way we dan easily change it on production if we run into conflicts.
```suggestion
        'prefix' => env('STATAMIC_SITEMAP_PREFIX, 'statamic_sitemap_'),
```

src/Jobs/GenerateCollectionSitemapJob.php
BobWez98: Agree

src/Jobs/GenerateCollectionSitemapJob.php
```diff
@@ -0,0 +1,56 @@
...
+            . $this->site->attribute('magento_store_id')
+            . '_collection_'
+            . $this->collection->handle()
+            . '.xml';
+
+        Storage::disk('public')->put($sitemapPath, $sitemapContent);
+    }
+
+
+    protected function generateCollectionSitemap() : string
```
BobWez98, line 44: Why not place this in the action?

src/Jobs/GenerateTermSitemapJob.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    protected GenerateSitemapAction $sitemapGenerator;
+
+    public function __construct(
+        protected Site $site,
+        protected Taxonomy $taxonomy,
+    ) {
+        $this->sitemapGenerator = new (GenerateSitemapAction::class);
```
BobWez98, line 27: Also here we can inject this

BobWez98 (changes requested):

kevinmeijer97:
This PR is now depending on https://github.com/rapidez/sitemap/pull/2 for it to generate the sitemaps. Although it's still possible to run the command manually, this PR would essentially schedule the sitemap generation to twice per day.

src/Jobs/GenerateCollectionSitemapJob.php
kevinmeijer97: I'd rather keep the logic for getting the collection entries in the job, same goes for getting the terms in the term job. The action itself is just for building the sitemap based on the given urls.

src/RapidezStatamicServiceProvider.php
kevinmeijer97: Changed this around so it hooks into https://github.com/rapidez/sitemap/pull/2.
See: https://github.com/rapidez/statamic/pull/78#issuecomment-2475852372

composer.json
```diff
@@ -31,15 +31,22 @@
     "prefer-stable": true,
     "require": {
         "php": "^8.1|^8.2",
+        "justbetter/statamic-glide-directive": "^2.1",
         "rapidez/blade-directives": "^0.6",
         "rapidez/core": "^2.13",
-        "justbetter/statamic-glide-directive": "^2.1",
+        "rapidez/sitemap": "@dev",
```
BobWez98, line 37: let's do 1.0.0

composer.json
```diff
@@ -31,15 +31,22 @@
...
         "statamic-rad-pack/runway": "^7.6",
         "statamic/cms": "^5.29",
         "statamic/eloquent-driver": "^4.9",
         "tormjens/eventy": "^0.8"
     },
+    "repositories": {
+        "rapidez/sitemap": {
+            "type": "path",
+            "url": "/Users/kevinmeijer/sites/rapidez-packages/sitemap"
+        }
```
BobWez98, line 48: This can be removed

src/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,24 @@
...
+use Rapidez\Statamic\Jobs\GenerateStoreSitemapJob;
+use Statamic\Facades\Site as SiteFacade;
+
+class GenerateSitemap extends Command
+{
+    protected $signature = 'rapidez:statamic:generate:sitemap';
+
+    protected $description = 'Generate a sitemap based on Statamic collections & taxonomies';
+
+
```
BobWez98, line 15: one line can be removed

BobWez98 (changes requested):

composer.json
```diff
@@ -31,9 +31,10 @@
     "prefer-stable": true,
     "require": {
         "php": "^8.1|^8.2",
+        "justbetter/statamic-glide-directive": "^2.1",
         "rapidez/blade-directives": "^0.6",
         "rapidez/core": "^2.13",
-        "justbetter/statamic-glide-directive": "^2.1",
+        "rapidez/sitemap": "^1.0",
```
royduin, line 37: This should be raised when https://github.com/rapidez/sitemap/pull/2 is merged and released.

src/Actions/GenerateSitemapAction.php
royduin, line 1: I think it's better that `rapidez/sitemap` provides this so `rapidez/statamic` can use it. Any other packages which needs to register a sitemap could also benefit from it.

src/Commands/GenerateSitemap.php
royduin, line 1: Maybe remove it and provide an option from `rapidez/sitemap`:

- `rapidez:sitemap:generate` = all
- `rapidez:sitemap:generate statamic`
- `rapidez:sitemap:generate anything-else`

And register new sitemaps with Eventy with a key so that can be used here.

src/Jobs/GenerateStoreSitemapJob.php
```diff
@@ -0,0 +1,66 @@
...
+                    })->isNotEmpty();
+            });
+
+        foreach ($taxonomies as $taxonomy) {
+            /* @var Taxonomy $taxonomy */
+            GenerateTermSitemapJob::dispatch($this->site, $taxonomy);
+        }
+    }
+
+    protected function addSitemap($url = null) : string
```
royduin, line 54: Also move this to `rapidez/sitemap` and use that here.

royduin (changes requested):

BobWez98:
Conflicts

src/RapidezStatamicServiceProvider.php
```diff
@@ -213,6 +219,42 @@ public function bootUtilities() : static
         return $this;
     }
 
+    public function bootSitemaps(): static
+    {
+        $storeId = Site::current()->attribute('magento_store_id');
+
+        // Cache the sitemaps for the specified store ID, refreshing every day
+        $sitemaps = Cache::remember(
```
royduin, line 227: This will also run with a composer install I think?

PR #79, opened 2024-10-07
Caching of menu and resolve urls of entries

BobWez98:
This Pull request contains a few things:
* PR #74 - Menu templates
* Caching of a navigation tag 
* resolving of entry urls

The parsing of urls will make sure if you select a category content, brand content and product content entry, the url will get resolved by either the normal statamic entry that's selected, of the runway resource that's linked to the collection. This way we only need one field in the navigation blueprint. This way we can also use the normal entry field in content components, and just use the helper to get the url. I also added some listeners to set the  settings of the navigation right when it's created, and clear the cache of the navigation when it's updated. 

Using the determineEntryUrl helper:
```php
RapidezStatamic::determineEntryUrl($entry)
```
This will return a string for the href attribute.

Using the nav helper to fetch the navigation:
```php
RapidezStatamic::nav('nav:main');
```
This will return an array of items with a resolved url.


composer.json
```diff
@@ -54,7 +55,10 @@
         "laravel": {
             "providers": [
                 "Rapidez\\Statamic\\RapidezStatamicServiceProvider"
-            ]
+            ],
+            "aliases": {
+                "StatamicNav": "Rapidez\\Statamic\\Facades\\StatamicNav"
```
indykoning, line 60: Do we want a facade this specific for `StatamicNav`? 
Maybe a facade for e.g. `RapidezStatamic` so any future functionalities can be added without having to create separate facades and possibly moving `StatamicNav` to that as well

resources/views/components/nav.blade.php
```diff
@@ -0,0 +1,51 @@
+@props(['nav' => 'nav:main', 'mobileNav' => 'nav:main'])
+@php
+    // This combines multiple navigations into one for desktop and mobile seperately, only neccessary if you have multiple navigations
+    $navData = array_merge(...Arr::map(Arr::wrap($nav), fn($item) => StatamicNav::nav($item)));
```
indykoning, line 4: These are not too readable, i suggest putting it in a Collection instead and using those functions.

src/Facades/StatamicNav.php
```diff
@@ -0,0 +1,16 @@
...
+use Illuminate\Support\Facades\Facade;
+
+/**
+ * @see \Rapidez\Statamic\StatamicNav
+ */
+class StatamicNav extends Facade
+{
+    protected static function getFacadeAccessor()
+    {
+        return 'rapidez-statamic-nav';
```
indykoning, line 14: I suggest returning `\Rapidez\Statamic\StatamicNav::class`
Makes it easier to find the correct class by simply clicking through which is not possible if it's a basic string

src/Listeners/ClearNavTreeCache.php
```diff
@@ -0,0 +1,21 @@
...
+
+class ClearNavTreeCache
+{
+    public function handle(NavTreeSaved $event): void
+    {
+        /** @var NavTree $tree */
+        $tree = $event->tree;
+        
+        $cacheKey = 'nav:' . $tree->handle() . '-' . config('rapidez.store');
+        if (Cache::has($cacheKey)) {
```
indykoning, line 17: I'm pretty sure this check isn't needed as forget already lets the respective driver check if it needs to remove it.
e.g. the file cache: https://github.com/laravel/framework/blob/5caf767f708aa6e8a3774104ecc7dd0471db805e/src/Illuminate/Cache/FileStore.php#L250

src/StatamicNav.php
```diff
@@ -0,0 +1,75 @@
...
+            $item['url'] = $this->determineEntryUrl($item['entry_id']->augmentable());
+            $branch[] = $item;
+        }
+
+        return $branch;
+    }
+
+    public function determineEntryUrl(Entry|Page $entry, string $nav = 'global-link'): string
+    {
+        return Cache::rememberForever(sprintf('%s-%s-nav-%s', config('rapidez.store'), $nav, $entry->id()), function () use ($entry) {
```
indykoning, line 50: I'm afraid we'll be creating a lot of caches, and retrieving a lot of caches by storing and retrieving this way.

We can reduce it by saving all urls into an array per `store-nav`
like done here
https://github.com/rapidez/core/pull/567/files#diff-061a7aa858d94ff473663e973e3097fbf1944640542b81be93f1961b42604bb3

however here we have a singleton and thus access to do 
`$this->navCache[$nav][$entry->id()] = $value`
instead of having to use both the array driver and regular cache driver

src/RapidezStatamicServiceProvider.php
```diff
@@ -70,6 +76,7 @@ public function bootCommands() : self
     public function bootConfig() : self
     {
         $this->mergeConfigFrom(__DIR__.'/../config/rapidez/statamic.php', 'rapidez.statamic');
+        $this->mergeConfigFrom(__DIR__ . '/../config/rapidez/statamic/nav.php', 'rapidez.statamic.nav');
```
indykoning, line 79: Don't forget to update your bootPublishables, since 2 new config files have been added https://github.com/rapidez/core/pull/356/files#diff-251f27d5c5089c981407ccb64c4235715c0e2021ac4765b7f1bfea2e7795e608R113

resources/views/components/nav.blade.php
BobWez98: Agreed

royduin (changes requested):

resources/views/navigation/header-button.blade.php
```diff
@@ -0,0 +1 @@
+{{-- This is shown on the first slideover of the mobile menu, this is empty by default but can be overridden --}}
```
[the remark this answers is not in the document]
JimmyHoenderdaal: Not really since this the left side of the navigation header, empty is fine by default
in some projects we add a search icon for the autocomplete slideover
image

src/RapidezStatamic.php
```diff
@@ -0,0 +1,85 @@
...
+                return $entry->url() ?? '';
+            }
+
+            $suffix = str($linkedRunwayResourceKey)->contains('category')
+                ? Rapidez::config('catalog/seo/category_url_suffix', '')
+                : (
+                    str($linkedRunwayResourceKey)->contains('product')
+                    ? Rapidez::config('catalog/seo/product_url_suffix', '')
+                    : ''
+                );
```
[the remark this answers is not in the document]
BobWez98: I've decided on a match here. Sadly this works by getting the entry from the navigation and then getting the runway resource. This runway resources returns as array and not as model, so we couldn't use any of the routing magic. On top of that the runway routing acts like actual routing, and we don't want that. We want the category routing from Magento.

src/RapidezStatamicServiceProvider.php
```diff
@@ -97,6 +103,9 @@ public function bootListeners() : self
                 Cache::forget('statamic-globals-' . Site::selected()->handle());
             });
 
+            Event::listen(NavCreated::class, SetCollectionsForNav::class);
```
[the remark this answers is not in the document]
BobWez98: No, only when a new single navigation is created, this will set the allowed collections on that new single navigation. This are just defaults and when you need or feel like changing it you're free to do so.

kevinmeijer97 (approved):

BobWez98:
I reversed the config files as discussed, any more feedback? @royduin @indykoning

kevinmeijer97 (approved):

indykoning (approved):

royduin:
I also merged this in master for v5 🚀

PR #80, opened 2024-10-17
Hybrid Runway setup

royduin:
In the early Statamic integration days we synced all products and categories from Magento into Statamic collections. This resulted in imports and a lot of data within collections, without the Eloquent driver this wasn't performing well.

The next step was the Eloquent driver integration and we came across Runway; so all the imports where removed and all products and categories where directly available within Statamic. But with that we still needed a place to store the additional data; for example the page builder fieldset on products and categories. So a "product content" and "category content" collection were created where the content was stored. With that a "Linked" field was introduced with a "Belongs To" relationship. With this you we're able to link to all products and categories through those relationships. If you'd like to enrich those with extra content from Statamic you'd to create a "content" collection and link it to the Runway resource.

At a moment we wanted all the content collections already there; so you don't have to create one and link it manually, with the result that all the import we're coming back. With that we're almost back to where we started.

---

So with this; a hybrid solution. Runway ~~all~~ half the way! On those blueprints you can have anything and when it's getting stored we're checking the changes and move those to entries. With that we're never touching the database as that should be read-only. No more imports and "content collections" for Runway resources.

- Currently only applied to products, but we should repeat this with categories and brands.
- We need something so we can easily link to Runway resources instead of having "Belongs To" relationships for all Runway resources.


src/Models/Product.php
```diff
@@ -20,10 +23,70 @@ protected static function booting()
...
         });
+
+        static::updating(function (Product $product) {
+            $productEntry = $product->entry;
+
+            if (!$productEntry) {
+                $productEntry = Entry::make()
+                    ->collection('products')
+                    ->locale(Site::selected()->handle())
+                    ->data(['linked_product' => $product->sku]);
```
BobWez98, line 34: Currently we link through `entity_id` , i think this needs to be kept in order to support backwards compatibility with content already configured in the old situation.

src/RapidezStatamicServiceProvider.php
```diff
@@ -136,7 +130,12 @@ public function bootComposers() : self
                     ->where('collection', 'products')
                     ->where('site', $this->getSiteHandleByStoreId())
                     ->where('linked_product', config('frontend.product.sku'))
-                    ->first();
+                    ->first()
+
+                    // As the "product content" collection is removed
+                    // with it's blueprint we need to specify the
+                    // runway blueprint, but this doesn't work.
+                    ->blueprint('runway/product');
```
BobWez98, line 138: We don't need the blueprint here as we're just querying the database for the entry, only important thing is to get the right entry and that's handled in the where clause above.

src/Models/Product.php
royduin: Nope, for products the SKU is used. Categories use the `entity_id`.

src/RapidezStatamicServiceProvider.php
royduin: When you're going to use the data on the frontend (with: https://github.com/rapidez/statamic#showing-content-on-categories-and-products) you'll end up with errors because the blueprint can't be found. When you put the old `product` blueprint here it's working fine.

kevinmeijer97: So it does also work with the `runway/product ` blueprint instead of the old `product` one?

royduin: Nope, as mentioned in the comment. Maybe the path is incorrect, didn't dive into this further.

kevinmeijer97:
There are some conflicts

UPGRADING.md
royduin, line 1: Let's move this to https://docs.rapidez.io/3.x/packages/statamic.html, when it's to much there we could add a collapse there.

composer.json
```diff
@@ -32,13 +32,14 @@
...
         "justbetter/statamic-glide-directive": "^2.1",
-        "rapidez/blade-directives": "^1.0",
+        "rapidez/blade-directives": "^0.6",
         "rapidez/core": "^2.13|^3.0",
         "rapidez/sitemap": "^1.1",
         "spatie/once": "*",
         "statamic-rad-pack/runway": "^7.6",
         "statamic/cms": "^5.29",
         "statamic/eloquent-driver": "^4.9",
+        "tdwesten/statamic-builder": "dev-main",
```
royduin, line 42: This can be changed to a specific version now.

composer.json
```diff
@@ -32,13 +32,14 @@
     "require": {
         "php": "^8.1|^8.2",
         "justbetter/statamic-glide-directive": "^2.1",
-        "rapidez/blade-directives": "^1.0",
+        "rapidez/blade-directives": "^0.6",
```
royduin, line 35: Why downgrading?

royduin:
Anything left? Is it ready to merge?

composer.json
BobWez98: Because when testing in rapidez 2.x, the core requires 0.x, 1.x throws an installation error.

royduin: It should work with both right?

royduin:
Let's merge when approved by @kevinmeijer97, @BobWez98 and @indykoning

indykoning (commented):
Don't forget to update and test the blade directive dependency, so it will be compatible with Rapidez v2 and v3.
Other than that, LGTM!

kevinmeijer97 (dismissed):
Agreed with https://github.com/rapidez/statamic/pull/80#pullrequestreview-2578701026

BobWez98:
I also added the 1.x version of the blade directives. 👍

BobWez98 (approved):

PR #81, opened 2024-10-17
Automate sites.yaml config

kevinmeijer97:
We can automate the generation of the new sites.yaml.
By doing this we reduce the amount of config needed to configure the sites, only keeping a few settings in the sites array of `config/rapidez/statamic.php`.

Not sure yet if setting this during boot is the way to go, could also generate it through a command when setting up the site.
Let me know what you would prefer!

### DIFF
--- a/config/rapidez/statamic.php
+++ b/config/rapidez/statamic.php
@@ -80,13 +80,10 @@ return [
 
     'sites' => [
         'default' => [
-            'name' => env('APP_NAME', 'Statamic'),
             'locale' => 'en_EN',
             'lang' => 'en_EN',
             'url' => '/',
             'attributes' => [
-                'magento_store_id' => 1,
-                'group' => 'default',
                 'disabled' => false,
             ],
         ],
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -42,6 +42,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
         $this
             ->bootCommands()
             ->bootConfig()
+            ->bootSites()
             ->bootRoutes()
             ->bootViews()
             ->bootListeners()
@@ -53,6 +54,32 @@ class RapidezStatamicServiceProvider extends ServiceProvider
 
         Vue::register();
         Alternates::register();
+    }
+
+    public function bootSites() : self
+    {
+        $sites = [];
+        $stores = Rapidez::getStores();
+
+        foreach ($stores as $store) {
+            $sites[$store['code']] = [
+                'name' => $store['name'] ?? $store['code'],
+                'locale' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.locale }}',
+                'lang' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.lang }}',
+                'url' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.url }}',
+                'attributes' => [
+                    'magento_store_id' => $store['store_id'],
+                    'group' => $store['website_code'] ?? '',
+                    'disabled' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.attributes.disabled }}',
+                ]
+            ];
+//            dd($store, $sites);
+        }
+
+        Site::setSites($sites);
+        Site::save();
+
+        return $this;
     }
 
     public function bootCommands() : self
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -73,7 +73,6 @@ class RapidezStatamicServiceProvider extends ServiceProvider
                     'disabled' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.attributes.disabled }}',
                 ]
             ];
-//            dd($store, $sites);
         }
 
         Site::setSites($sites);
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -2,7 +2,6 @@
 
 namespace Rapidez\Statamic;
 
-use Statamic\Statamic;
 use Statamic\Sites\Sites;
 use Statamic\Facades\Site;
 use Statamic\Facades\Entry;
@@ -33,7 +32,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
     public function register()
     {
         $this->app->extend(Sites::class, function () {
-            return new SitesLinkedToMagentoStores(config('statamic.sites'));
+            return new SitesLinkedToMagentoStores();
         });
     }
 


src/RapidezStatamicServiceProvider.php
```diff
@@ -55,6 +55,31 @@ public function boot()
...
+    {
+        $sites = [];
+        $stores = Rapidez::getStores();
+
+        foreach ($stores as $store) {
+            $sites[$store['code']] = [
+                'name' => $store['name'] ?? $store['code'],
+                'locale' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.locale }}',
+                'lang' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.lang }}',
+                'url' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.url }}',
```
royduin, line 68: We can also get this from Magento right? The locales and url's are also configured there?

src/RapidezStatamicServiceProvider.php
```diff
@@ -55,6 +55,31 @@ public function boot()
         Alternates::register();
     }
 
+    public function bootSites() : self
+    {
+        $sites = [];
+        $stores = Rapidez::getStores();
```
royduin, line 61: We shouldn't query anything directly from the Service Provider, as it's executed even from your console when you didn't have a database connection yet. But if we move it to a middleware it's only executed on browser requests; not when you run something in the CLI. But this generates the sites config file? Maybe a command?

royduin (changes requested):

src/Commands/CreateSites.php
```diff
@@ -0,0 +1,46 @@
...
+                'attributes' => [
+                    'magento_store_id' => $store['store_id'],
+                    'group' => $store['website_code'] ?? '',
+                    'disabled' => '{{ config:rapidez.statamic.sites.' . $store['code'] . '.attributes.disabled }}',
+                ]
+            ];
+        }
+
+        Site::setSites($sites);
+        Site::save();
```
BobWez98, line 42: Since we're getting this from the database anyway, can't we maybe just overwrite the `getSavedSites` function in `Statamic\Sites\Sites` to just return a cached array with the queries already done? Then we set the sites on the fly and we can get rid of the sites.yaml all together 🤔

BobWez98 (commented):
This feels a bit manual, maybel we can add some magic to do it on the fly, but this is just food for thought.

src/Commands/CreateSites.php
kevinmeijer97: Yes i agree, i feel like this might be the best way to go.
I cached the sites array for now because of the setStore() function which is needed to get the correct site data.
After https://github.com/rapidez/core/pull/567 gets merged we can use getValue() function without having to do the setStore().

src/Commands/CreateSites.php
```diff
@@ -0,0 +1,46 @@
+<?php
```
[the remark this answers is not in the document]
kevinmeijer97: Yes good catch, explained further in https://github.com/rapidez/statamic/pull/81#discussion_r1806395996

BobWez98 (approved):

src/Extend/SitesLinkedToMagentoStores.php
```diff
@@ -20,4 +22,38 @@ public function findByMageRunCode($code)
...
+                $url = $configModel::getCachedByPath('web/secure/base_url');
+
+                $sites[$store['code']] = [
+                    'name' => $store['name'] ?? $store['code'],
+                    'locale' => $locale,
+                    'lang' => $lang,
+                    'url' => $url,
+                    'attributes' => [
+                        'magento_store_id' => $store['store_id'],
+                        'group' => $store['website_code'] ?? ''
```
royduin, line 51: Previously the group was explained in the readme, with this PR that info will be removed. Maybe we should add some info on the alternate hreflang link tags in the readme?

src/Extend/SitesLinkedToMagentoStores.php
```diff
@@ -20,4 +22,38 @@ public function findByMageRunCode($code)
...
+                    ]
+                ];
+            }
+
+            return $sites ?: $this->getFallbackConfig();
```
royduin, line 56: The method was changed recently: https://github.com/statamic/cms/pull/10470, do we require the versions where that's changed? https://github.com/statamic/cms/releases/tag/v5.16.0

src/RapidezStatamicServiceProvider.php
```diff
@@ -33,7 +32,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
     public function register()
     {
         $this->app->extend(Sites::class, function () {
-            return new SitesLinkedToMagentoStores(config('statamic.sites'));
+            return new SitesLinkedToMagentoStores();
         });
```
royduin, line 36: This can be rewritten to a oneliner now?

royduin (changes requested):

config/rapidez/statamic.php
```diff
@@ -80,13 +80,10 @@
 
     'sites' => [
```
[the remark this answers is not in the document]
kevinmeijer97: Changed the readme for this!

src/Extend/SitesLinkedToMagentoStores.php
kevinmeijer97: Added some info for this into the readme

src/Extend/SitesLinkedToMagentoStores.php
```diff
@@ -20,4 +22,38 @@ public function findByMageRunCode($code)
...
+            foreach ($stores as $store) {
+                if (config('rapidez.statamic.sites.' . $store['code'] . '.attributes.disabled')) {
+                    continue;
+                }
+
+                Rapidez::setStore($store['store_id']);
+
+                $locale = $configModel::getCachedByPath('general/locale/code');
+                $lang = explode('_', $locale)[0] ?? '';
+                $url = $configModel::getCachedByPath('web/secure/base_url');
```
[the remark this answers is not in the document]
kevinmeijer97: This happened to be some missing Rapidez config in the Magento repo. The .env of the Magento repo missed the Rapidez urls which led to the `web/secure/base_url` config being the Magento url instead of the Rapidez url.

kevinmeijer97:
I think we should wait with merging so we can release it in bulk along with the Hybrid runway changes.

src/RapidezStatamicServiceProvider.php
```diff
@@ -32,9 +31,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
 {
     public function register()
     {
-        $this->app->extend(Sites::class, function () {
-            return new SitesLinkedToMagentoStores(config('statamic.sites'));
-        });
+        $this->app->extend(Sites::class, fn () => new SitesLinkedToMagentoStores());
```
BobWez98, line 34: Isn't `$this->app->extend(Sites::class, SitesLinkedToMagentoStores::class);` enough?

BobWez98 (dismissed):

README.md
```diff
@@ -24,7 +24,9 @@ composer require rapidez/statamic
 ## Install command
 
 The install command will help you set up all the necessary settings.
-It will mainly setup the [Eloquent driver](https://github.com/statamic/eloquent-driver) and publish the necessary vendor from the rapidez/statamic repo. 
+It will mainly setup the [Eloquent driver](https://github.com/statamic/eloquent-driver) and publish the necessary vendor from the rapidez/statamic repo.
+
+Before running the install script, make sure you have an existing User model as Statamic requires this. See: [Statamic docs](https://statamic.dev/tips/storing-users-in-a-database)
```
royduin, line 29: This doesn't explain how you should get you user model back: https://github.com/laravel/laravel/blob/11.x/app/Models/User.php, as we don't provide it from `rapidez/rapidez`: https://github.com/rapidez/rapidez/blob/040ada124619473f7d713aab537d82cea51f0026/.github/workflows/copy-laravel-source.yml#L45. You'll also need the migrations, etc.

kevinmeijer97:
Moved most of the configuration guide to the Rapidez docs (https://github.com/rapidez/docs/pull/56).

src/RapidezStatamicServiceProvider.php
kevinmeijer97: No the second parameter had to be a closure, keeping it on this for now

config/rapidez/statamic.php
```diff
@@ -80,13 +80,10 @@
...
         'default' => [
-            'name' => env('APP_NAME', 'Statamic'),
             'locale' => 'en_EN',
             'lang' => 'en_EN',
             'url' => '/',
             'attributes' => [
-                'magento_store_id' => 1,
-                'group' => 'default',
                 'disabled' => false,
             ],
```
BobWez98, line 88: Do we still need this whole sites config?

src/Extend/SitesLinkedToMagentoStores.php
```diff
@@ -20,4 +22,38 @@ public function findByMageRunCode($code)
...
+
+    protected function getSavedSites()
+    {
+        return Cache::rememberForever('statamic_sites', function () {
+            $sites = [];
+            $stores = Rapidez::getStores();
+            $configModel = config('rapidez.models.config');
+
+            foreach ($stores as $store) {
+                if (config('rapidez.statamic.sites.' . $store['code'] . '.attributes.disabled')) {
```
BobWez98, line 34: Can't we safely say we always want this to be enabled? And otherwise maybe just have an array that includes the disabled stores?

BobWez98 (dismissed):

BobWez98:
We can merge this in my opinion, what about yours @royduin

royduin:
We've a conflict but besides that I'm also curious about the answers on your latest questions. It's currently unclear where the `sites` in the config is for or when to use it.

kevinmeijer97:
> We've a conflict but besides that I'm also curious about the answers on your latest questions. It's currently unclear where the `sites` in the config is for or when to use it.

Conflict has been fixed!
Also refactored the sites array and replaced it with a `disabled_sites` array, there is no need to have a sites array in the config anymore when the only thing configurable is the disabled option.
You can add the store_codes of the stores you don't want in Statamic to the `disabled_sites` array.

src/Extend/SitesLinkedToMagentoStores.php
kevinmeijer97: Changed this around, see https://github.com/rapidez/statamic/pull/81#issuecomment-2457255941

config/rapidez/statamic.php
kevinmeijer97: Changed this around, see https://github.com/rapidez/statamic/pull/81#issuecomment-2457255941

README.md
kevinmeijer97: Changed the readme with a clearer explanation for how and what with the user model!

BobWez98 (dismissed):

config/rapidez/statamic.php
```diff
@@ -78,17 +78,7 @@
...
-            'attributes' => [
-                'magento_store_id' => 1,
-                'group' => 'default',
-                'disabled' => false,
-            ],
-        ],
-    ]
+    'disabled_sites' => [
+        // '{site_handle}'
+    ],
```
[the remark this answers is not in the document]
kevinmeijer97: Changed the docs for this: https://github.com/rapidez/docs/pull/62

src/Extend/SitesLinkedToMagentoStores.php
```diff
@@ -22,6 +24,40 @@ public function findByMageRunCode($code)
         if (!$code || !($this->sites instanceof Collection)) {
             return null;
         }
-        return $this->sites->get($code));
+        return $this->sites->get($code);
+    }
+
+    protected function getSavedSites()
+    {
+        return Cache::rememberForever('statamic_sites', function () {
```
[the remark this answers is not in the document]
kevinmeijer97: Changed the docs for this: https://github.com/rapidez/docs/pull/62

BobWez98 (approved):

PR #84, opened 2024-10-25
Use the Statamic web middleware

royduin:
This depends on: https://github.com/rapidez/core/pull/616. Instead of repeating that again I'm just registering the middleware here again so the FallbackController can use it.

We need this for the live preview. Without it it's not working.


royduin:
Better approach: https://github.com/rapidez/statamic/pull/89

PR #89, opened 2024-11-14
Load original statamic fallback controller

indykoning:
This PR supersedes: #84

Why do we want to do this? 
It has no impact on current functionality, we're just cleaning some code up that we don't need.

However with: rapidez/core/pull/617 this will ensure the middlewares are loaded for this controller,
thus making the preview function, as one of the many other benefits functional!


BobWez98 (approved):

royduin:
Could you fix the conflict 😇

indykoning:
The conflicts have been resolved 🙂

PR #90, opened 2024-11-18
Added default for missing User model

kevinmeijer97:
The added User model is the complete User model with the Statamic changes added to it as well.


src/RapidezStatamicServiceProvider.php
```diff
@@ -186,6 +189,18 @@ public function bootPublishables() : self
             __DIR__.'/../config/rapidez/statamic.php' => config_path('rapidez/statamic.php'),
         ], 'config');
 
+        if (!class_exists('\App\Models\User::class') && $userModel = $this->prepareUserModel()) {
```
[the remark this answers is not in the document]
kevinmeijer97: Yes i agree, this was not the way to go.
Changed it so that it's loaded from the module now!

BobWez98 (dismissed):

src/RapidezStatamicServiceProvider.php
```diff
@@ -189,6 +192,21 @@ public function bootPublishables() : self
         return $this;
     }
 
+    public function bootUserModel() : static
+    {
+        if (!class_exists('\App\Models\User::class')) {
+            config(['auth.providers.users.model' => User::class]);
+        }
+
+        if (!Schema::hasTable('users') && !file_exists(database_path('migrations/0001_01_01_000000_create_users_table.php'))) {
```
royduin, line 201: You don't want to have database actions within the service provider as this runs even when you didn't have this setup yet for example with composer

src/Models/User.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * The attributes that should be hidden for serialization.
+     *
+     * @var array<int, string>
+     */
+    protected $hidden = [
+        'password',
+        'remember_token',
+    ];
+
```
royduin, line 33: No docblocks here, maybe just remove all the docblocks from the model to keep it inline?

src/RapidezStatamicServiceProvider.php
kevinmeijer97: Removed the hasTable check, it's not really needed for this anyways. Checking if the migration file exists is enough.

BobWez98 (approved):

database/migrations/0001_01_01_000000_create_users_table.php
royduin, line 1: Instead keeping a copy within this repo we could get the content from https://github.com/laravel/laravel/blob/11.x/database/migrations/0001_01_01_000000_create_users_table.php? Maybe just a `file_get_contents()` on the publish?

royduin (commented):
What about these steps? https://statamic.dev/tips/storing-users-in-a-database#in-an-existing-laravel-app, do we also need to do something with that here?

kevinmeijer97:
> Instead keeping a copy within this repo we could get the content from https://github.com/laravel/laravel/blob/11.x/database/migrations/0001_01_01_000000_create_users_table.php? Maybe just a `file_get_contents()` on the publish?

Changed this around to get a copy from the repo instead.

> What about these steps? https://statamic.dev/tips/storing-users-in-a-database#in-an-existing-laravel-app, do we also need to do something with that here?

These steps are in the docs at this moment, only doing this automatically right now because the user model + migration is needed for the install of this addon.

BobWez98:
Conflicts

src/RapidezStatamicServiceProvider.php
```diff
@@ -199,6 +203,31 @@ public function bootPublishables() : self
...
+        }
+
+        $migrationPath = database_path('migrations/0001_01_01_000000_create_users_table.php');
+
+        if (!file_exists($migrationPath)) {
+            try {
+                $userTable = file_get_contents('https://raw.githubusercontent.com/laravel/laravel/refs/heads/11.x/database/migrations/0001_01_01_000000_create_users_table.php');
+
+                if (!$userTable) {
+                    throw new Exception('Failed to download the migration file.');
```
BobWez98, line 219: So if laravel moves this file, we'll get exceptions on every production site that includes this code? 😲

BobWez98 (changes requested):

src/RapidezStatamicServiceProvider.php
kevinmeijer97: No, as this will be caught and logged in the catch.
Also will only happen if there is no existing migration, which should never be on a production environment.

BobWez98 (approved):

royduin:
I'm not really sure about this PR, downloading from a boot method; having the user model within here as fallback. Maybe we've to talk this through later.

BobWez98:
> I'm not really sure about this PR, downloading from a boot method; having the user model within here as fallback. Maybe we've to talk this through later.

Agreed

PR #91, opened 2024-11-29
Static caching with invalidation on Magento urls

royduin:
It just adds the static caching middleware on the `web` middleware group, this way all urls can be cached. To invalidate changed pages the `rapidez-statamic:invalidate-cache` should be scheduled, all info will go into the docs later.

There are some todo's left:

- Docs
- Double check with multisites
- Optionally; cache warming

Ah and.. we could cherry pick this into 4.x if we want this in Rapidez v2 installations.

### DIFF
--- a/src/Commands/InvalidateCacheCommand.php
+++ b/src/Commands/InvalidateCacheCommand.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace Rapidez\Statamic\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\DB;
+use Rapidez\Core\Facades\Rapidez;
+use Statamic\StaticCaching\Cacher;
+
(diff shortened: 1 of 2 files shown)


src/Commands/InvalidateCacheCommand.php
```diff
@@ -0,0 +1,126 @@
...
+        foreach ($stores as $store) {
+            Rapidez::setStore($store);
+
+            $this->urls = collect();
+
+            $this
+                ->addProductsUrls()
+                ->addCategoryUrls()
+                ->addPageUrls()
+                ->makeAbsolute();
```
BobWez98, line 44: Isn't this just invalidating all the urls too? Why not do `$cacher->flush();` here too then? Or am i missing something?

BobWez98:
Maybe a todo for later; invalidate cache through a magento webhook on an update.

BobWez98 (approved):

kevinmeijer97 (approved):

PR #92, opened 2024-12-05
Use whereJsonContains

Jade-GG:
The `linked_product` and `linked_category` fields are a "Belongs To" field which is one of those fields that always has `max_items:1`. Normally, when you would save an entry, this field would just return the number, like so:

```js
{
    ...
    linked_category:92
    ...
}
```

However, since a semi-recent update of runway, if you save the blueprint, it will silently remove "max_items:1" from the field, and you end up getting this:

```js
{
    ...
    linked_category:[92]
    ...
}
```

This is sort of a bug, or at least an unintended side effect of some update.

---

This is a backwards-compatible way to make sure this update doesn't break your links. Whether or not this is the ideal solution, I'm not sure. Ideally this bug should just get fixed of course 🙂

### DIFF
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -149,7 +149,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
                 $entry = Entry::query()
                     ->where('collection', 'products')
                     ->where('site', $this->getSiteHandleByStoreId())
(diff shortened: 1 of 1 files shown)


BobWez98 (approved):

royduin:
Let's see where https://github.com/statamic-rad-pack/runway/issues/640 goes

PR #95, opened 2024-12-12
Get site by Store ID fix

kevinmeijer97:
In some cases there is a problem in the way the `magento_store_id` attribute in the sites.yaml is being parsed through antlers.
When getting the attribute from the site the store ID is not always being cast as an integer but as a string.

Screenshot 2024-12-12 at 10 52 14

I've made some changes to cast this as an integer.


indykoning (dismissed):

BobWez98 (dismissed):

royduin:
Just replacing `===` with `==` seems cleaner? Not very readable at the moment.

PR #97, opened 2024-12-16
Added Attribute + Attribute Option Runway Models

kevinmeijer97:
Added 2 new Runway Models for Attributes & Attribute Options.
This will be used for the Query Builder which will be added in another PR.


src/Models/ProductAttribute.php
```diff
@@ -0,0 +1,212 @@
...
+
+class ProductAttribute extends Model
+{
+    use HasRunwayResource;
+
+    protected $table = 'eav_attribute';
+    
+    protected $primaryKey = 'attribute_id';
+
+    protected $fillable = [
```
royduin, line 21: It's read only so we don't need this?

src/Models/ProductAttribute.php
```diff
@@ -0,0 +1,212 @@
...
+        $optionValues = array_filter(explode(',', $this->option_values));
+
+        if (count($optionIds) !== count($optionValues)) {
+            return [];
+        }
+
+        return array_combine($optionIds, $optionValues);
+    }
+
+    public function getFormattedOptionsAttribute()
```
royduin, line 130: The new accessor syntax is cleaner: https://laravel.com/docs/master/eloquent-mutators#defining-an-accessor

src/Models/ProductAttribute.php
```diff
@@ -0,0 +1,212 @@
...
+            ]);
+        });
+    }
+
+    protected static function getCurrentStoreId(): string
+    {
+        return once(fn() => (Statamic::isCpRoute()
+            ? (Site::selected()->attributes['magento_store_id'] ?? '1')
+            : (Site::current()->attributes['magento_store_id'] ?? '1')
+        ));
```
royduin, line 111: Maybe extract this somewhere as it's duplicated among multiple models now.

src/Models/ProductAttribute.php
```diff
@@ -0,0 +1,172 @@
...
+        });
+    }
+
+   public function options(): Attribute
+   {
+       return Attribute::make(
+           get: fn () => !$this->option_ids || !$this->option_values ? [] : 
+               (count($optionIds = array_filter(explode(',', $this->option_ids))) === count($optionValues = array_filter(explode(',', $this->option_values))) ? array_combine($optionIds, $optionValues) : [])
+       );
+   }
```
BobWez98, line 87: Maybe this one could be more readable when we use collections and their functions?

```php
    return Attribute::make(
        get: fn () => collect([$this->option_ids, $this->option_values])
            ->map(fn($value) => collect(explode(',', $value))->filter())
            ->pipe(function (Collection $collections) {
                $optionIds = $collections[0];
                $optionValues = $collections[1];

                if ($optionIds->count() === $optionValues->count()) {
                    return $optionIds->combine($optionValues);
                }

                return [];
            })
    );
```

src/Models/ProductAttribute.php
```diff
@@ -0,0 +1,172 @@
...
+
+   public function formattedOptions(): Attribute
+   {
+       return Attribute::make(
+           get: fn () => empty($this->options) ? '' : 
+               collect($this->options)
+                   ->map(fn($value, $id) => "{$value} (ID: {$id})")
```
BobWez98, line 94: Can't we use single quotes here?

src/Models/ProductAttribute.php
```diff
@@ -0,0 +1,172 @@
...
+   public function frontendLabel(): Attribute
+   {
+       return Attribute::make(
+           get: fn () => $this->store_frontend_label ?? $this->attributes['frontend_label'] ?? ''
+       );
+   }
+
+    public function getCacheKey(): string
+    {
+        return "product_attribute_{$this->attribute_id}_store_" . static::getCurrentStoreId();
```
BobWez98, line 108: And here

src/Models/ProductAttribute.php
```diff
@@ -0,0 +1,172 @@
...
+    {
+        parent::boot();
+
+        static::saved(function ($model) {
+            Cache::forget($model->getCacheKey());
+        });
+
+        static::deleted(function ($model) {
+            Cache::forget($model->getCacheKey());
+        });
```
BobWez98, line 170: We could put this into an observer to clean this file up a bit, also you don't have to overwrite the boot method https://laravel.com/docs/11.x/eloquent#observers

src/Models/ProductAttributeOption.php
```diff
@@ -0,0 +1,129 @@
...
+            get: fn () => $this->attributes['store_value'] ?? $this->admin_value
+        );
+    }
+
+    public function DisplayValue(): Attribute
+    {
+        return Attribute::make(
+            get: fn () => $this->store_value ?: $this->admin_value
+        );
+    }
```
royduin, line 83: They should still begin with a lowercase.

src/Models/ProductAttributeOption.php
```diff
@@ -0,0 +1,129 @@
...
+            ]);
+        });
+    }
+
+    public function attribute()
+    {
+        return $this->belongsTo(ProductAttribute::class, 'attribute_id', 'attribute_id');
+    }
+
+    public function AdminValue(): Attribute
```
BobWez98, line 64: Do we want this to start with uppercase character? the docs say: 

```php
public function adminValue(): Attribute
```

would equal to 
```php
$model->admin_value
```

src/Models/ProductAttributeOption.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        return Attribute::make(
+            get: fn () => $this->attributes['admin_value'] ?? ''
+        );
+    }
+
+    public function StoreValue(): Attribute
```
BobWez98, line 71: Same here

src/Models/ProductAttributeOption.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        return Attribute::make(
+            get: fn () => $this->attributes['store_value'] ?? $this->admin_value
+        );
+    }
+
+    public function DisplayValue(): Attribute
```
BobWez98, line 78: And here

src/Models/ProductAttributeOption.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        parent::boot();
+
+        static::saved(function ($model) {
+            Cache::forget($model->getCacheKey());
+        });
+
+        static::deleted(function ($model) {
+            Cache::forget($model->getCacheKey());
+        });
```
BobWez98, line 127: I think `ProductAttribute` and `ProductAttributeOption` can use the same observer :)

src/Models/ProductAttribute.php
```diff
@@ -0,0 +1,172 @@
...
+                'eav_attribute_label.value'
+            ]);
+        });
+    }
+
+   public function options(): Attribute
+   {
+       return Attribute::make(
+           get: fn () => !$this->option_ids || !$this->option_values ? [] : 
+               (count($optionIds = array_filter(explode(',', $this->option_ids))) === count($optionValues = array_filter(explode(',', $this->option_values))) ? array_combine($optionIds, $optionValues) : [])
```
royduin, line 85: This isn't very readable; maybe better to not force this on 1 line

src/Models/ProductAttributeOption.php
```diff
@@ -0,0 +1,129 @@
...
+
+    public function scopeByAttributeCode($query, $attributeCode)
+    {
+        return $query->where('eav_attribute.attribute_code', $attributeCode);
+    }
+
+    public function scopeOrderBySort($query)
+    {
+        return $query->orderBy('eav_attribute_option.sort_order');
+    }
```
BobWez98, line 115: Do we need these scopes? As we're only modifying 1 thing on the query?

BobWez98 (changes requested):

src/Models/ProductAttribute.php
kevinmeijer97: I'd rather keep it like this, keep it more compact

src/Models/ProductAttribute.php
kevinmeijer97: I'd rather keep it like this, keep it more compact

src/Models/ProductAttributeOption.php
kevinmeijer97: Added these first because i figured we might need these for the Query Builder.
Looks like i won't be needing it so far, removing them for now!

src/Observers/ModelObserver.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Rapidez\Statamic\Observers;
+
+use Illuminate\Support\Facades\Cache;
+use Rapidez\Statamic\Models\ProductAttribute;
+use Rapidez\Statamic\Models\ProductAttributeOption;
+
+class ModelObserver
```
BobWez98, line 9: This name implicates it observes all models. I think "CacheObserver" or something is a more describing name

src/Models/ProductAttributeOption.php
```diff
@@ -0,0 +1,101 @@
...
+
+class ProductAttributeOption extends Model
+{
+    use HasRunwayResource;
+
+    protected $table = 'eav_attribute_option';
+
+    protected $primaryKey = 'option_id';
+
+    public $timestamps = false;
```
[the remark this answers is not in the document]
kevinmeijer97: Yes, as the default of this is true. Which is not needed

royduin:
Merged it so we can move forward. Any other feedback can go in a new PR.

A little disclaimer; I did not review the queries and the solution direction. **So everything is subjected to change later!** For example; I'm not sure if we want these big queries and complexity as with Rapidez we're moving to a more "Eloquent relations way". An example of that is how the product options are handled: https://github.com/rapidez/core/blob/master/src/Models/ProductOption.php, furthermore all the scopes are currently unused, there is an observer on read-only models, etc.

royduin:
+ Maybe this needs to move to https://github.com/rapidez/statamic-query-builder as it's not used from this package.

PR #99, opened 2024-12-24
[4.x] Reduce queries

BobWez98:
This PR adds some Statamic overwrites. It disables the hydration of globals as we're doing that ourselves. It also puts the meta query's for assets in the Cache. Sadly Statamic uses this logic in the control panel so we do need to execute it. By putting it in the Cache store instead of in Blink it reduces the queries on the frontend hugely.

Before these changes:
Scherm­afbeelding 2024-12-24 om 10 35 25

After these changes:
Scherm­afbeelding 2024-12-24 om 10 36 33

I'll put this as a draft to have it open for discussion, we can add this for 5.x later when necessary.

### DIFF
--- a/src/Eloquent/Assets/AssetRepository.php
+++ b/src/Eloquent/Assets/AssetRepository.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Rapidez\Statamic\Eloquent\Assets;
+
+use Illuminate\Support\Facades\Cache;
+use Statamic\Contracts\Assets\Asset as AssetContract;
+use Statamic\Support\Str;
+use Statamic\Eloquent\Assets\AssetRepository as StatamicAssetRepository;
+
+class AssetRepository extends StatamicAssetRepository
+{
+    public function findById($id): ?AssetContract
+    {
+        [$container, $path] = explode('::', $id);
+
+        $filename = Str::afterLast($path, '/');
+        $folder = Str::contains($path, '/') ? Str::beforeLast($path, '/') : '/';
+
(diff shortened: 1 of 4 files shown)


src/Eloquent/Assets/AssetRepository.php
```diff
@@ -0,0 +1,29 @@
...
+    public function findById($id): ?AssetContract
+    {
+        [$container, $path] = explode('::', $id);
+
+        $filename = Str::afterLast($path, '/');
+        $folder = Str::contains($path, '/') ? Str::beforeLast($path, '/') : '/';
+
+        $cacheKey = "eloquent-asset-{$id}";
+
+        return Cache::rememberForever($cacheKey, function () use ($container, $filename, $folder) {
```
indykoning, line 21: Do we want this remembered forever? are there any potential issues doing that, getting stale data?
Or do relevant changes update the id in this case?

If we see issues we could potentially use flexible or a cache with an expiry

BobWez98:
I decided to remove the part I extended the Asset fieldtype + repository. This part is needed for statamic to get the data of the asset field and should not be cached like this. It also doesn't really matter if get the information from cache or from the database. With the new static caching functionalities it should run once on the frontend anyway. It doesn't really affect the performance much.

Globals however are redundant for our implementation. I kept the overwrite and i suggest we keep it.

royduin:
Maybe a description within the code where it's for?

BobWez98:
Done!

BobWez98:
> Do we also need this on master / 5.x?

https://github.com/rapidez/statamic/pull/113 :)

PR #102, opened 2025-01-02
[5.x] Automaticly set the config of the static caching paths when enabled

BobWez98:
If we're automaticly setting the statamic sites we can also set the static caching paths. This way we don't need the urls in the `.env` when using static caching.


royduin:
Do we need to mention this somewhere in the docs?

PR #109, opened 2025-01-13
Do startPush on view composer instead of boot

Jade-GG:
Alternative of: https://github.com/rapidez/core/pull/701

The issue arises when we get an error while attempting to render a page, e.g. a 404. Here's a play by play of what happens:

1. A user navigates to a nonexistent page that will throw a 404
2. The app boots
3. The pipelines and controllers all happen, eventually arriving at Statamic's frontendcontroller which tries to find a page
4. Statamic renders a view, fails, and throws a 404
5. The Laravel Pipeline catches this error and goes on to render a 404 page. Critically, this doesn't re-boot the app because it has already booted.

Note that both step 4 and step 5 render a view, and also note that when a view gets rendered, it will flush the stacks. Ultimately, this means that when you get a 404 page, this startPush has been consumed by Statamic and will not actually push to the final page render.

*EDIT: Note also that there may definitely be other cases where more than one view gets rendered per boot. I couldn't give you any concrete examples, but I would not be surprised if there are legitimate use cases for that.*

---

The above PR to the Rapidez core fixes the stack flush at step 4 by tricking Laravel into expecting a JSON response.  
The PR you're looking at right now makes it so the stack gets pushed to whenever the view is rendered.

### DIFF
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -242,9 +242,7 @@ class RapidezStatamicServiceProvider extends ServiceProvider
 
     public function bootStack() : static
     {
-        if (! $this->app->runningInConsole()) {
-            View::startPush('head', view('statamic-glide-directive::partials.head'));
-        }
+        View::composer('rapidez::layouts.app', fn($view) => $view->getFactory()->startPush('head', view('statamic-glide-directive::partials.head')));
 
         return $this;
     }


royduin:
So this vs https://github.com/rapidez/core/pull/701? What's the best option and why?

Jade-GG:
> So this vs [rapidez/core#701](https://github.com/rapidez/core/pull/701)? What's the best option and why?

My personal take is that solving it like this is the cleanest option. This is because the core version really only attacks a specific cause of the bug that might not catch everything, and it does so in a kinda hacky way by tricking laravel into responding with a json under specific conditions.

However, having both of these merged would probably be nice either way. Stopping Laravel from unnecessarily rendering an extra view is always positive.

kevinmeijer97:
I think the problem with this solution is that it only fixes it for the startPush of the glide directive view.
It would still break if another package makes use of the startPush, the solution in https://github.com/rapidez/core/pull/701 would fix this problem.
That's why i'd personally go for the other solution.

Jade-GG:
> I think the problem with this solution is that it only fixes it for the startPush of the glide directive view. It would still break if another package makes use of the startPush, the solution in [rapidez/core#701](https://github.com/rapidez/core/pull/701) would fix this problem. That's why i'd personally go for the other solution.

If anything, doesn't that just show that doing `View::startPush()` like this is just not how you should do it? We already needed the `runningInConsole()` exception because of the weird place this is in (it might be a good idea to show the correct way in the Rapidez docs)

Again it's not like the 404 from statamic is the only way this issue could possibly happen.

Jade-GG:
We could also provide a helper function in the Rapidez facade to do this.

PR #117, opened 2025-02-21
Updated statamic/cms dependency

kevinmeijer97:


royduin:
https://github.com/rapidez/statamic/pull/111 requires https://github.com/statamic/cms/pull/11381 which was released 5.47.0

PR #122, opened 2025-03-11
Added custom cache invalidator

kevinmeijer97:
Added a basic cache invalidator for emptying the caches when the relevant content has changed.
This new class can also be extended and used as a base in projects where more custom validation rules are required.


src/RapidezStatamicServiceProvider.php
```diff
@@ -249,6 +251,17 @@ public function bootSitemaps(): static
...
+    public function bootStaticCaching(): static
+    {
+        if (config('statamic.static_caching.invalidation.class')) {
+            return $this;
+        }
+        
+        config()->set('statamic.static_caching.invalidation.class', CustomInvalidator::class);
+
+        return $this;
+    }
```
BobWez98, line 263: Why not add this in the config file? then it can be overwritten too, now it's not that clear that it's possible to overwrite this class

kevinmeijer97: I don't feel like we should make some config for this, since it's already basic Statamic config.
I'd rather add some more documentation for this in https://docs.rapidez.io/3.x/packages/statamic.html.

BobWez98 (approved):

indykoning (approved):

src/RapidezStatamicServiceProvider.php
royduin, line 263: Maybe flip it around so there is one return:

```suggestion
    public function bootStaticCaching(): static
    {
        if (!config('statamic.static_caching.invalidation.class')) {
            config()->set(
                'statamic.static_caching.invalidation.class',
                CustomInvalidator::class
            );
        }
        
        return $this;
    }
```

src/StaticCaching/CustomInvalidator.php
```diff
@@ -0,0 +1,36 @@
...
+
+        $urls = [];
+
+        if ($item instanceof Entry && $item->collectionHandle() === 'categories' && $item->linked_category) {
+            $urls[] = Str::ensureLeft($item->linked_category['url_path'], '/');
+        }
+
+        if ($item instanceof Entry && $item->uri()) {
+            $urls[] = $item->uri();
+        }
```
royduin, line 27: This will cover products and brands? Isn't this already covered by the default invalidator?

royduin (approved):
But let's wait with the merge until we also have a PR to master

kevinmeijer97:
[5.x] version here: https://github.com/rapidez/statamic/pull/127

PR #124, opened 2025-03-18
Save latestCheck with static files

indykoning:
This fixes the LatestCheck from being deleted by clearing the cache.

Instead of selecting an alternative caching method i've opted with using the same methods as statamic uses for static caching.
This has the following advantages:
- We cannot accidentally select a cache driver that will also be cleared
- The latest execution is visible plaintext when logging into the server
- The latestCheck value will only be removed when clearing the entire static cache (using the backend or the statamic comment)

When approved i will also create a PR for master


royduin:
Looking good to me! 🚀 Let's merge when the 5.x PR is there.

PR #126, opened 2025-04-04
[5.x] Use the retrieved event instead of having to overwrite getAttributes

BobWez98:
I've tested this with all the types (Brands, Categories and Products). This will add the right attributes to the model and render them in the entry in the backend of Statamic. This won't change anything for the frontend. It's still adding the content to the views as expected.

I also updated Runway. There's been some fixes for relation fields between Runway models. If you have a relations field on a Runway model called "products" it expected to have a "products" relation on the runway model. This is fixed now.

The upgrade guide show's no high impacting changes that applicable for us: https://runway.duncanmcclean.com/upgrade-guides/v7-to-v8

### DIFF
--- a/config/rapidez/statamic.php
+++ b/config/rapidez/statamic.php
@@ -49,19 +49,16 @@ return [
             \Rapidez\Statamic\Models\Product::class => [
                 'name' => 'Products',
                 'title_field' => 'name',
-                'cp_icon' => 'table',
             ],
 
             \Rapidez\Statamic\Models\Category::class => [
                 'name' => 'Categories',
                 'title_field' => 'name',
-                'cp_icon' => 'array',
             ],
 
             \Rapidez\Statamic\Models\Brand::class => [
                 'name' => 'Brands',
                 'title_field' => 'value_store',
-                'cp_icon' => 'tags',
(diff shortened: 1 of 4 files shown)


config/rapidez/statamic.php
Jade-GG, line 1: How come these icons are removed?

src/Observers/RunwayObserver.php
```diff
@@ -41,13 +41,25 @@ public function updating(Model $model)
         return false;
     }
 
+    public function retrieved(Model $model)
+    {
+        $originalAttributes = $model->getAttributes();
+        
+        if ($model->exists && $modelEntry = ($model->entry ?? '')) {
+            $model->setRawAttributes(array_merge(
+                $modelEntry->data()->toArray(),
```
Jade-GG, line 50: Might be cleaner to just do `$model->entry->data()->toArray()` and skip setting a variable within the if statement.

Jade-GG (changes requested):

config/rapidez/statamic.php
BobWez98: see "low impact changes" https://runway.duncanmcclean.com/upgrade-guides/v7-to-v8

PR #133, opened 2025-04-30
[5.x] InvalidateCacheCommand multistore compatible

BobWez98:
This PR adds store code to static cache path when we have multiple paths defined, even though there might only be one store, the developer still could've set these config values as an array.

Tested on 4.x and 5.x, with multistore and single store.

4.x: #134

### DIFF
--- a/src/Commands/InvalidateCacheCommand.php
+++ b/src/Commands/InvalidateCacheCommand.php
@@ -42,7 +42,7 @@ class InvalidateCacheCommand extends Command
                 $this->info('Cleared all urls (as we do not have a latest check date yet)');
                 $cacher->flush();
                 $this->setLatestCheckDate($writer);
-                return;
+                continue;
             }
             $this->setLatestCheckDate($writer);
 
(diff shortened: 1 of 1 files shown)


src/Commands/InvalidateCacheCommand.php
```diff
@@ -20,22 +20,31 @@ class InvalidateCacheCommand extends Command
...
+
+            $this->staticCachePath = config($staticCachePathConfig);
+
+            $this->latestCheck = $this->getLatestCheckDate();
+
+            if (!$this->latestCheck) {
+                $this->info('Cleared all urls (as we do not have a latest check date yet)');
+                $cacher->flush();
+                $this->setLatestCheckDate($writer);
+                return;
```
indykoning, line 45: Since we're within a foreach now, wouldn't we want to use continue in case other stores do have a latestcheck file?

BobWez98: 🔪

indykoning (approved):

PR #136, opened 2025-05-02
Submit forms with fetch

BobWez98:
This adds a version with the translations being passed as prop. I will make a second PR whith a proposal for merging translations with Rapidez Core.

Also see #137 for a better handling of translations

### DIFF
--- a/resources/js/components/FormConditions.vue
+++ b/resources/js/components/FormConditions.vue
@@ -3,22 +3,59 @@ import '/public/vendor/statamic/frontend/js/helpers.js'
 Vue.prototype.Statamic = window.Statamic
 
 export default {
-    props: ['initialData'],
+    props: {
+        initialData: {
+            type: Object,
+            required: true,
+        },
+        translations: {
+            type: Object,
+            default: () => ({}),
+        },
+        callback: {
+            type: Function,
+        },
+    },
     data() {
         return {
             formData: this.initialData,
+            success: false,
+            error: false,
         };
     },
     render() {
-        return this.$scopedSlots.default({ formData: this.formData });
+        return this.$scopedSlots.default({ formData: this.formData, submit: this.submit });
     },
     mounted() {
+        console.log(this.translations)
         let token = this.$root.csrfToken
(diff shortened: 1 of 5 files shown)


resources/js/components/FormConditions.vue
```diff
@@ -3,22 +3,59 @@ import '/public/vendor/statamic/frontend/js/helpers.js'
...
+    props: {
+        initialData: {
+            type: Object,
+            required: true,
+        },
+        translations: {
+            type: Object,
+            default: () => ({}),
+        },
+        callback: {
```
BobWez98, line 15: Added a callback because in a project you might want to add some functionality like "closing a modal" or conversion tracking.

resources/js/components/FormConditions.vue
```diff
@@ -3,22 +3,59 @@ import '/public/vendor/statamic/frontend/js/helpers.js'
...
+            type: Function,
+        },
+    },
     data() {
         return {
             formData: this.initialData,
+            success: false,
+            error: false,
```
BobWez98, line 23: The developer can use these to show/hide validation on the frontend apart from the notify.

Jade-GG:
I agree that #137 is nicer when the translations thing is merged 🙂

resources/js/components/FormConditions.vue
```diff
@@ -19,6 +33,28 @@ export default {
...
+
+            if (response.ok) {
+                this.success = true
+                Notify(this.translations.success, 'success')
+                if (this.callback) {
+                    await this.callback()
+                }
+            } else {
+                this.error = true
+                Notify(this.translations.error, 'error')
```
indykoning, line 55: In some places in Rapidez we also log the error https://github.com/rapidez/core/blob/master/resources/js/fetch.js#L150
https://github.com/rapidez/core/blob/4ddd4d2c1fd0a3f6ae0f2dea81331a0a5ef7254a/resources/js/stores/useSwatches.js#L21
is that something we might want to do here?

BobWez98:
As discussed this will be closed.

PR #139, opened 2025-05-09
Set staticCachePathConfig in foreach

jbclaudio:


src/Commands/InvalidateCacheCommand.php
```diff
@@ -25,9 +25,9 @@ class InvalidateCacheCommand extends Command
     public function handle(Cacher $cacher, Writer $writer): void
     {
         $stores = Rapidez::getStores();
-        $staticCachePathConfig = 'statamic.static_caching.strategies.full.path';
 
         foreach ($stores as $store) {
+            $staticCachePathConfig = 'statamic.static_caching.strategies.full.path';
```
BobWez98, line 30: What does this change? it is the same inside as outside the foreach right?

BobWez98 (approved):

PR #144, opened 2025-06-27
[4.x] Added Attribute + Attribute Option Runway Models

kevinmeijer97:
This is the 4.x version for https://github.com/rapidez/statamic/pull/97.
Didn't think this was needed at first, but it would be nice to have this PR available in 4.x as well.
The changes are also non breaking so it should be good to go.


kevinmeijer97:
No i agree this is not the right place for this anymore, it should be moved to the Query Builder instead.
I'm closing this and creating a ticket for the Query Builder.

PR #148, opened 2025-07-18
[5.x] Hack in namespace so statamic gets the correct blueprint

Jade-GG:
When getting the entry directly like here, Statamic doesn't find the right blueprint because it only looks in the `collections/[handle]` folder. This is hardcoded within Statamic so there's no easy way around that. This means that all category pages and product pages in the frontend (not the backend) currently break due to there being no blueprint attached to them, giving you raw data instead of augmented data.

However, we can trick statamic into looking elsewhere by setting a custom namespace like this. It's a pretty gross hack, I don't like it. But it's the only thing I found that works.

We have to make sure it doesn't run when in the backend because otherwise it will show the extra blueprints from these hacky namespaces.

### DIFF
--- a/src/RapidezStatamicServiceProvider.php
+++ b/src/RapidezStatamicServiceProvider.php
@@ -175,6 +175,11 @@ class RapidezStatamicServiceProvider extends ServiceProvider
             )]);
         }
 
+        if (! Statamic::isCpRoute()) {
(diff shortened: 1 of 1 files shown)


BobWez98 (approved):

royduin:
This wasn't an issue earlier, maybe something changed in Runway? Maybe better to fix it there?

BobWez98:
I'd say merge this and create a follow-up story to research other solutions.

PR #153, opened 2025-08-06
Added option for ajax response on Statamic forms

kevinmeijer97:
In some cases, you don't want to redirect after form submission, but rather stay on the current page and handle the success message based on the successful response.

By default, Statamic form submission endpoints return a 302 response unless you specify the request with an `XMLHttpRequest` header.

This approach allows you to tell the response to return a 200 status with JSON data instead.


resources/js/components/FormConditions.vue
```diff
@@ -11,10 +11,15 @@ export default {
...
         },
+        ajaxResponse: {
+            type: Boolean,
+            default: false
+        }
     },
     data() {
         return {
             formData: this.initialData,
+            ajaxResponse: this.ajaxResponse,
```
BobWez98, line 22: This is not needed, if it's a prop you can just use it as this.ajaxResponse without declaring it in the data

BobWez98 (dismissed):

indykoning (approved):

PR #157, opened 2025-08-21
Create dictionary form field type

Jade-GG:
(which is just the select type but fancy)


Jade-GG:
> I'd say just use the select component instead, since it's the same anyways. Seems a bit project specific to add this too

it is an actual Statamig field that can be used, for things like country lists.

kevinmeijer97:
Yeah i understand, but might as well use the existing select for it. Check for it in the form template instead

BobWez98:
If it works with the `@include` i think adding the template for it is a good addition, no reason to add duplication of course. Maybe some screenshots of why you need a dictionary?

royduin:
+ keep in mind we do have v5 (for Rapidez v3) and v6 (for Rapidez v4) now!

Jade-GG:
> Yeah i understand, but might as well use the existing select for it. Check for it in the form template instead

As we don't currently have any checks in the form already, I'd rather not start making the template ugly (and very slightly making the update require updating any overwritten templates) just to add this one field. Instead I've turned it into an `include` for now to remove duplication, which seems cleaner to me anyway. Especially if you want to actually modify what a dictionary looks like without modifying the select (for whatever reason).

kevinmeijer97 (approved):

BobWez98 (approved):

PR #159, opened 2025-08-28
[6.x] Allow for arbitrary link keys instead of just the primary key

Jade-GG:
This is useful for when you don't want to change your primary key, which could have various reasons.


royduin:
Any more info on this?

Jade-GG:
> Any more info on this?

This is specifically used for product attributes in one of our ongoing projects, because the primary key is `attribute_id` but we would rather use `attribute_code` as the link key. This PR allows us to do exactly that.

PR #181, opened 2026-02-13
[8.x] Minimize memory usage for retrieving product rewrite URLs

Jade-GG:
See #180


royduin:
This should also go into 7.x right?

PR #184, opened 2026-02-16
[7.x] Fix response typehint in boot

Jade-GG:
See rapidez/core#1194


royduin:
This should also go into 7.x right?

PR #187, opened 2026-03-31
[6.x] Use brand collection instead of Runway model

kevinmeijer97:
Brands is the only model that allows for having custom (Statamic) routes, the problem lies with saving the slug when you are on the edit page of a model entry.

Since we added the collection blueprint on the model page in kind of a hacky way it cant build the slug field correctly when set, it always tries to base it of the title instead of getting the slug from the database.

To fix this we should try and stay closer to Statamic by moving the edit to the collection.
I've set the model to hidden, as we do need the model for the import.
Then i've set the collection back to visible, so its available through the menu.

Also changed the default fields to a page builder top and bottom, as we do this manually in each project now anyways.
I will make a PR for 7.x later.

### DIFF
--- a/config/rapidez/statamic.php
+++ b/config/rapidez/statamic.php
@@ -63,6 +63,7 @@ return [
                 'name' => 'Brands',
                 'title_field' => 'value_store',
                 'order_by' => 'sort_order',
+                'hidden' => true,
             ],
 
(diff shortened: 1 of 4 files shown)


BobWez98 (approved):

indykoning (approved):
Is this a breaking change? And if so, should we write it down somewhere?

PR #189, opened 2026-04-24
Added migration code and pipelines to migrate Magento pages to Statamic

indykoning:
This PR adds the ability to migrate HTML into Statamic bard fields.
As an example (since every Magento and Statamic installation has them) a cms pages to statamic pages migration command has been added using this migration.

Making use of pipelines for most of the HTML transformations so the migration and transformation process can be customized to a customer's needs.

e.g. currently iframes are transformed into urls, since Statamic does not support iframes in bard.
If a user has a fieldset for iframes, the `TransformIframes` could be removed. And a custom step for turning it into an iframe fieldset could be used instead.


kevinmeijer97 (approved):

royduin:
Let's wait for the Statamic v6 PR, when that's merged we can merge this one.

PR #201, opened 2026-06-08
[8.x] Also flush application cache on global flush

Jade-GG:
Without this, things like the navigation menu will not get invalidated properly. As we can't perfectly predict which caches are necessary to clear (including, but not limited to, use of `@includeCached`), the only good solution is to flush the entire cache.


royduin:
Do we also need this in older versions?

PR #202, opened 2026-06-08
[6.x] Open details accordions when there are errors

Jade-GG:
ref: BORDEX-1410

This issue arose in a certain project because there were accordions inside of the form, but they couldn't be focused or shown to have an error because they were hidden in a closed accordion. This event listener opens these accordions (recursively, in case there are multiple parent accordions) so that the field is shown.


royduin:
We also need this on `7.x` and `master`?

Jade-GG:
> We also need this on `7.x` and `master`?

Correct, I hadn't gotten to that yet.

PR #208, opened 2026-07-06
Fix replicators not working from Runway resource blueprints

Jade-GG:
ref: AN-432

Since what I assume would be the Statamic 6 upgrade, the `vendor.runway` hint path is used for the blueprint token in Replicators and Bards. This path does not get recognized because of the way we have set up our Runway integration.

This means that instead of using the correct path `runway::category` using the runway namespace, it will spit out the raw `vendor.runway.category` and look for a blueprint with that exact name.

This line gives Statamic enough information to not do this, and look for the blueprint in the proper location.


kevinmeijer97 (approved):

royduin:

Jade-GG:
> Do we need this in other versions?

It wasn't an issue before we updated the relevant project to Statamic 6 so I don't think so? However, I haven't actually checked to make sure. I'll make an issue to look at this.

PR #209, opened 2026-07-06
Support non-integer key types

Jade-GG:
ref: AN-432

We ran into an issue where Runway entries using UUIDs didn't get saved properly. Turns out it was trying to cast the UUID into an integer while saving because it's the primary key, because Laravel automatically casts primary keys to whatever the defined `keyType` is.

By adding `string` as the `keyType` here we allow any type of key to be used.

### DIFF
--- a/src/Models/BaseEntry.php
+++ b/src/Models/BaseEntry.php
@@ -14,6 +14,7 @@ class BaseEntry extends Model
     use Searchable;
 
     protected $table = 'statamic_entries';
+    protected $keyType = 'string';
     protected static $collection;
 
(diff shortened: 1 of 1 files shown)


BobWez98 (approved):

kevinmeijer97 (approved):

royduin:

Jade-GG:
> Do we need this in other versions?

It wasn't an issue before we updated the relevant project to Statamic 6 so I don't think so? However, I haven't actually checked to make sure. I'll make an issue to look at this.

PR #210, opened 2026-07-07
Added form fields checkboxes and radio's

Roene-JustBetter:
When selecting checkboxes or radio from the Statamic form builder those elements where not showing on the frontend because these files where missing here.


royduin:

PR #211, opened 2026-07-10
Update formfields added extra checks for possible empty display fields

Roene-JustBetter:
- Added labels above radio's and checkboxes. 
- Added checks for `$field['display']` this could be left empty in Statamic because it is not a required field. 

_Maybe nice to have but we could discuss this later option to show instructions field on the frontend below or above the field_

### DIFF
--- a/resources/views/form_fields/checkboxes.blade.php
+++ b/resources/views/form_fields/checkboxes.blade.php
@@ -1,11 +1,15 @@
+@if ($field['display'] ?? false)
+    <x-rapidez::label>@lang($field['display'])</x-rapidez::label>
+@endif
 @foreach($field['options'] as $key => $label)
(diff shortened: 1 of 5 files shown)


kevinmeijer97 (approved):

resources/views/form_fields/checkboxes.blade.php
```diff
@@ -1,11 +1,15 @@
+@if ($field['display'] ?? false)
+    <x-rapidez::label>@lang($field['display'])</x-rapidez::label>
```
royduin, line 2: The label isn't clickable currently right?

Roene-JustBetter: Correct this is the label that is above the checkboxes: 
Image

PR #213, opened 2026-07-29
[8.x] Extended link to include Product and Category Runway models

kevinmeijer97:
Added an extend to the link field to register the Product and Category Runway models so they can be used in the link field.

Normally Runway models would automatically be registered, but since the Product and Category models aren't your default Runway models this was not possible so we had to register and resolve the urls like this instead.

Screenshot 2026-07-29 at 14 35 20


BobWez98 (dismissed):

indykoning (dismissed):

royduin:
Related to:
- https://github.com/statamic-rad-pack/runway/discussions/271
- https://github.com/statamic-rad-pack/runway/pull/838
- https://github.com/statamic/cms/pull/15004

---

- Do we still need the `determineEntryUrl()`?
- Do we need to raise the versions of Statamic and Runway in the `composer.json` as this is feature was recently added?
- With this we can just use `$item['url']` everywhere? We don't need things like this:
```
<a href="{{ isset($item['product']->value()['url']) ? '/'.$item['product']->value()['url'] : $item['url'] }}">
```

kevinmeijer97:
- Yes, for now we still need `determineEntryUrl()` as it is used for nav trees. Changing this to work with the models is possible but would create some breaking changes. If we really want that i'd suggest to create a separate PR for this, some type of migration script would be nice to have here as well. The same goes for `Can we just use $item['url'] everywhere?`
- I bumped the versions to `"statamic/cms": "^6.26"` &  `"statamic-rad-pack/runway": "^9.6"`

